PR #32, opened 2014-09-26
CONTRIBUTING.md

dominictarr:
Added a contribution document that basically explains what we have already been doing...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,19 @@
+## So you want to help with _secure-scuttlebutt_?
+
+_secure-scuttlebutt_ is a do-ocracy. that means ruled by doers.
+If you want to help, you need to figure out what you need to do and then do it.
+There are no managers. The best way to do this is probably to discuss with other
+contributors. for this use issues and irc (find us in #stackvm on irc)
+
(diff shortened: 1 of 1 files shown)


dominictarr:
@pfraze what do you think, this PR is out of date, and not quite what we have ended up doing,
we should probably what down what we do, so that newcomers can easily contribute.

pfraze:
here are my thoughts --

correctness is important in the ssb repo. we should encourage PRs, but you should approve them before merge. i think it'd be a bad idea to let people override your vote here

i'm not sure about the other repos. i'm not sure how the ssbc org is going to develop. i think it'd be a bad idea to let somebody become a maintainer on merit of contributions only. if they started making bad merges i'd be pretty upset. i'd want it to be somebody i could trust and hold accountable, and who communicates well. we hold each other to pretty high standards on this; same would appy to new people

maintainers dont get to be dictators, but they need to be responsible people that move the project forward

pfraze:
a lot of that is already reflected in the doc. i like stating a "path to contributorship," but i really would only want to give merge rights to people we know will work with the org

dominictarr:
Hmm, well one thing we should definitely mention is that ALL changes are made via pull requests, except for releases. This serves both to explain to other contributors what is going on, and also to enable discussion of changes.

PR #53, opened 2014-10-30
added automatic de-indexing with the withdraws reltype

pfraze:
https://github.com/dominictarr/secure-scuttlebutt/issues/42

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,7 +106,6 @@ module.exports = function (db, opts) {
 
       if (link.$rel == 'withdraws' && isHash(link.$msg)) {
         // remove from indexes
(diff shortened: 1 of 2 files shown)


dominictarr:
I don't want to merge this yet, because it's possible to get inconsistent indexes if it crashes before the get completes. It would be better to leave a feature out, than to have a feature that works unreliably*,
what we need to do before we merge this is have a way to get reliable async indexes.

\* This probably works nearly all the time, but that isn't good enough for a system that aspires to be as foundational as ssb.

That said, this would also have flow on effects to other datastructures that are represented across multiple messages... I think we need to aquire more experience with those (and possibly, represent deletes closer to the application layer...)

I'm gonna close this issue for now, but we should revisit this later.

Issue #76, opened 2015-01-11
standardizing getter output


pfraze:
in phoenix code, i've started to standardize the message objects in the `{key:, value:}` form, for two reasons: 

 1. i can modify that wrapper object without modifying the content of the message, so i can stick on new data (eg `msg.wasRead = true`)
 2. it gives me access to the msg's key

problem is that the getters are inconsistent on the output format and sometimes won't let you get it in this form

 - `createFeedStream` cant get it with keys
 - `createHistoryStream` cant get it with keys
 - `createLogStream` can only get it with keys
 - `messagesByType` respects `keys` opt
 - `messagesLinkedToMessage` cant get it with keys
 - all other link-based getters use their own format

so i'm thinking we should update the first 3 of those to respect the `keys` opt, and then update `messagesLinkedToMessage` to work like the other link-based getters

that sound good?

pfraze:
there's a similar problem with `add`, but with a different cause

muxrpc only supports callbacks with `(err, value)` and `add` does `(err, value, key)` so the key doesnt get sent over the wire. that's a problem for some of my code right now (i need the key)

for `add` i propose we either
 - change the cb to `(err, {key:, value:})` - my preference, but possibly more work
 - change muxrpc to support multiple args in the callback

dominictarr:
This is a good idea.
I'd prefer add to cb `{key, value}`. there might be one or two plugins that use this,
but we can change them.

pfraze:
ok, i'll make the changes in this branch

pfraze:
> and then update `messagesLinkedToMessage` to work like the other link-based getters

looking more deeply, it makes sense that `messagesLinkedToMessage` is different than the other link-based getters. i updated it to have a `keys:` opt instead

pfraze:
done

PR #77, opened 2015-01-11
standardizing getter output

pfraze:
see https://github.com/ssbc/secure-scuttlebutt/issues/76

started by adding `keys: true` support to `createFeedStream`. waiting on the rest till we have agreement

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,9 +256,12 @@ module.exports = function (db, opts) {
       paramap(function (data, cb) {
         var key = data.value
         var seq = data.key
-        db.get(key, function (err, value) {
-          cb(err, {key: key, value: value, timestamp: seq})
-        })
+        if (opts.keys)
+          db.get(key, function (err, value) {
+            cb(err, {key: key, value: value, timestamp: seq})
+          })
+        else
+          db.get(key, cb)
       })
     )
   }
--- a/index.js
+++ b/index.js
@@ -155,17 +155,38 @@ module.exports = function (db, opts) {
     })
   }
 
+  // opts standardized to work like levelup api
+  function stdopts (opts) {
+    opts = opts || {}
+    if (opts.keys !== false)
+      opts.keys = true
+    if (opts.values !== false)
+      opts.values = true
+    return opts
+  }
+  function msgFmt (keys, values, obj) {
+    if (keys && values)
+      return obj
+    if (keys)
+      return obj.key
+    if (values)
+      return obj.value
+    return null // i guess?
+  }
+
   //TODO: eventually, this should filter out authors you do not follow.
   db.createFeedStream = function (opts) {
-    opts = opts || {}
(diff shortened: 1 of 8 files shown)


pfraze:
@dominictarr ready for review

there were other places where the change could be made (eg `getLatest`) but i chose to leave that for now

pfraze:
updates should be minimal. only added the `keys: true` opts to most. `add()` change was breaking

pfraze:
oh, other breaking change - `createLogStream` no longer defaults to `keys: true` behavior, must be specified

dominictarr:
discussion in irc:

``` 
<domanic> I'm guessing you made everything consistent, and that is why log stream is different now?
<domanic> also for these lines: https://github.com/ssbc/secure-scuttlebutt/pull/77/files#diff-168726dbe96b3ce427e7fedce31bb0bcR259
<domanic> it would be more consistent with the rest of secure-scuttlebutt style to use a ternary
<domanic> _keys ? {key, value} : value
<domanic> or at least, put the if else inside the cb to get, instead of duplicating the get
<pfraze> k. and yeah thats why log stream changed
<domanic> pfraze, maybe we should just make this consistent with levelup?
<pfraze> default it to true and add the value opt?
<domanic> https://github.com/rvagg/node-levelup/blob/master/lib/levelup.js#L403-L418
<domanic> I'm not sure whether we'd need the values: false
<pfraze> yeah, i'm good w/that
<domanic> but being consistent is good
<domanic> hmm...
<domanic> though, since these are indexes do we really need the key, would it be better to use the index key?
<domanic> I'm really not sure
<domanic> maybe the index could be on another property, like ts is on log stream?
<domanic> {values, keys, indexes} ?
<pfraze> yeah that could work, i just need the key attr
<domanic> hmm, we should probably leave that until we actually have a case... oh yeah, we do have a case because that is on log stream
<domanic> although, it doesn't give the entire key, just the significant part (in that case, the timestamp)
<pfraze> if we use gt/lt/gte/lte then the index attr would be useful
<domanic> yeah. we should use that, and also we should support live streams
<domanic> oh yeah, I added that to the link streams recently
<domanic> pfraze, hmm, maybe we are over thinking this? 
<domanic> maybe we should just tidy up the if else and merge?
```

pfraze:
after sleeping on it i do feel like having `{keys:, values:}` behavior like in leveldb (where each can be by itself and they default to true) would be a good idea, because otherwise we're so similar. that would cause `createLogStream` to default behave as it did before, but `createFeedStream`, `createHistoryStream`, and `messagesLinkedToMessage` would change

it'll be a pain to change, but easier now than later?

dominictarr:
okay if you wanna go that way I'm :+1:

pfraze:
@dominictarr ok ready for review

pfraze:
published `8.0.0`

Issue #74, opened 2015-01-04
reject too-big messages


dominictarr:
oops, @pfraze created message `H6yP9cn3T7D4lB/Yf9ndjVD9gRPwBFImhjsldFtS1+0=.blake2s` which is way too and should not have been validated.

now there are two problems: fix validation to disallow such messages. clean the current instances so that invalid messages are removed.

pfraze:
yeah lets just discard that (my) feed and plan not to do that again

dominictarr:
what should have happened: the messages before that should have been accepted, but that messages (and thus any messages after that) should be rejected.

pfraze:
looking at scuttlebot's replication, it's a pretty straight pipe from the peer's history stream into a local write stream. is there validation occurring somewhere?

pfraze:
oh yes -- in the `.add` function

pfraze:
this has been handled

PR #78, opened 2015-01-18
validation now rejects messages larger than 1024 bytes

pfraze:
https://github.com/ssbc/secure-scuttlebutt/issues/74

is there a better way to do this? would be great not to do a throwaway `JSON.stringify`

### DIFF
--- a/test/validation.js
+++ b/test/validation.js
@@ -141,6 +141,29 @@ module.exports = function (opts) {
 
 
   })
+
+  tape('too big', function (t) {
+    var keys = opts.keys.generate()
+    var id = opts.hash(keys.public)
+    var prev
+    ssb.add(
+      prev = create(keys, null, {type: 'init', public: keys.public}),
+      function (err) {
+        if(err) throw explain(err, 'init failed')
+
+        var str = ''
+        for (var i=0; i < 101; i++) str += '1234567890'
+        ssb.add(
+          prev = create(keys, 'msg', str, prev),
+          function (err) {
+            if(!err) throw 'too big was allowed'
+            console.log(err)
+            t.end()
+          }
+        )
+      }
+    )
+  })
 }
 
 if(!module.parent)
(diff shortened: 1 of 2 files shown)


pfraze:
added a solution for https://github.com/ssbc/secure-scuttlebutt/issues/79

pfraze:
i'd need to put in some logging to confirm this, but i suspect that, even though (with this change) validation rejects too large messages and any subsequent messages, replication will continue to fetch latest from those feeds with the rejected messages, and so the same bad messages will get exchanged and rejected indefinitely

if so, we should create a "dead-feed" jail so that feeds with errors (such as a too-big msg, or a fork in the history) won't get replicated anymore. we should make sure, however, to only put feeds in the jail if we've proven the authorship of the erroneous message(s). in the event of a bad message, we'd do a signature check before jailing. otherwise, an attacker could put other users' feeds into jail by forging bad messages during replication.

pfraze:
do we want to consider something larger than `1kb`? like maybe `5kb`? what's the largest we could safely do?

dominictarr:
well, it's arbitary... packets are 1500 bytes. I guess 5kb would be okay. when I picked 1024 bytes we where using a binary format. Now 2kb might be a better option.

Ultimately, if the messages are larger, we just need to (soft-)enforce a lower messages/day limit.
(we talked about this once - peers just refuse to replicate more than a given number of messages in a day) so that peers cannot spam too much. This would work best as a shared rule across everything in the network, otherwise some nodes might end up permanently behind you, if they have a lower setting than everyone else does.

pfraze:
what prompted my size-increase comment was thinking about publishing a bundle of assets and realizing that 1kb can only contain about 20 links. if fitting in a packet makes tfx more efficient, though, that might be worth keeping.

dominictarr:
@pfraze what if we parsed the ext links and detected recursive links? we could have a json type, with more links, or links embedded in markdown css or html. or we you could just have a json manifest.

Or ssb could be aware of tarballs, and you could attach an entire directory (or we could use some tarball like bundle format that isn't so retro ;)

pfraze:
@dominictarr yeah, those are all good solutions for us to explore

PR #81, opened 2015-01-21
relatedMessages

dominictarr:
Basically the same as phoenix's `getThread`, except as a query instead of as a view/index.
It is simple to rebuild this on demand because we already have an index of links within messages, so it's just a few calls to get the "replies".

By passing in a `rel` type, you get only messages that reply by that type. Also supported are `count` and `parent` options, but they are off by default.

This can be used to implement comment threading, but also many other types of recursive structures based on message links.

I added api documentation, although I notice that the other documentation is kinda out of date.


pfraze:
looks good :+1:

dominictarr:
okay I'm gonna merge this.

PR #85, opened 2015-02-03
fix: failed feed.add calls no longer put the feed into a bad state

pfraze:
after merging https://github.com/ssbc/secure-scuttlebutt/pull/78, i found that after ssb rejects a `feed.add` (eg because it's too large) the subsequent calls to `feed.add` would be rejected with `seq too high`. this appears to be because `feed.add` doesnt reset some state when a message is rejected. this PR will fix


pfraze:
@dominictarr https://github.com/ssbc/secure-scuttlebutt/commit/01cc2d7c231a8d8cf185feea5a931dde66e435e8 solves this bug, but it does so by making `feed.add` wait for `ssb.add` to finish before queueing the new message. this means it's probably not taking advantage of the validator's processing pipeline as much as it could, so i wanted you to review this choice before i merged into this PR

feel free to branch this PR (so you'll have the failing test) and implement an alternative fix

dominictarr:
hmm, I think the solution is just to validate the length before it gets to to main validator, when doing local writes (add)

pfraze:
i had the same thought. there may be other validations that could create this situation, though, which can't be moved up. i'll look into it.

dominictarr:
true, but if it's really important, I think the user will wait for their add to succeed. being able to queue adds like this is really just a convenience for tests...

Actually... the reason I made it write in batches is so it could receive lots of messages during replication,
writing messages written by the local user is really not a big deal. just waiting until one has been successfully written before writing the next is completely okay.

pfraze:
i think to do this right we'd need to run the entire `validateSync` function before going into the rest of the validaiton pipeline, right? if any of `validateSync` checks fail, the feed will end up in a bad state

pfraze:
@dominictarr is the solution in https://github.com/ssbc/secure-scuttlebutt/commit/01cc2d7c231a8d8cf185feea5a931dde66e435e8 ok for you?

dominictarr:
sorry I missed this. what does https://github.com/ssbc/secure-scuttlebutt/commit/01cc2d7c231a8d8cf185feea5a931dde66e435e8 do?

pfraze:
I explained it up in https://github.com/ssbc/secure-scuttlebutt/pull/85#issuecomment-72707531, basically it makes the add function wait for each add to complete before running the next because if any of the validations fail, the feed will enter a bad state.

dominictarr:
I just realized that this only changes tests, but the tests are currently passing... so does this need to be merged?

pfraze:
oh right, yes this PR only changes tests, so we can merge it at any time. The tests in here show the issue so they should fail if we merge.
https://github.com/ssbc/secure-scuttlebutt/compare/post-rejection-state-solution1 includes my proposed solution, which i put in a separate branch in case there was a better way to solve it

pfraze:
I revisited this and realized that https://github.com/ssbc/ssb-feed/pull/3 is the only adequate solution because `ssb-feed` may be run in a process that communicates with the ssb-db over RPC, and so it needs to wait for each message to write successfully before updating its `prev` cursor.

@dominictarr good to merge both? This PR is just the tests, the ssb-feed PR is the fix.

dominictarr:
okay merged into 10.3.2

PR #90, opened 2015-02-06
fix: validation pipeline now drains after a validation error

pfraze:
We were experiencing long delays in the `feed.add()` function for reasons unknown (https://github.com/ssbc/phoenix/issues/299#issuecomment-73179046). Turns out it was because the validation pipeline was failing to resume processing after a message was rejected. Adding some `drain()` calls solved the issue.

Closes https://github.com/ssbc/phoenix/issues/299


pfraze:
published `8.1.3`

PR #95, opened 2015-02-27
tracking pull request: mynosql

dominictarr:
use generic queries instead of hardcoded.
There is more work to do here, so won't be merging this just yet.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,8 @@ var assert    = require('assert')
 var ltgt      = require('ltgt')
 var mlib      = require('ssb-msgs')
 var explain   = require('explain-error')
+
+var mynosql   = require('mynosql')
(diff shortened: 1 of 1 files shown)


dominictarr:
hey can we merge this? mynosql is pretty good now... and I need this for my talk in a few days.

pfraze:
ok I just spent some time looking at mynosql and how you use it in bittodo. it is looking good!

could it work without being integrated into ssb? eg in bittodo could you do `nomysql(rpc.ssb)`?

dominictarr:
@pfraze hmm, bittodo doesn't actually touch the db directly, it uses the client protocol.
that would work if bittodo had it's own server. (if we went down that road)

dominictarr:
merged into 10.1.0

Issue #94, opened 2015-02-22
A more constrained link definition in message objects


pfraze:
per discussion on irc, 
The `rel` attribute on links seems burdensome when we can just put links in sub-objects and infer the rel-type from the attribute name. I believe @dominictarr feels the same.

We discussed -and I now propose -we restrict links from appearing in the top object, and remove the `rel` attribute instead. A sub-object will then be determined to be a link by looking for `feed`, `msg`, and `ext`, which may be used in any combination on the sub-object. The relation type will be the key of the sub-object in its immediate parent.

For example, this message --

```json
  "content": {
    "type": "post",
    "text": "hello @dominic",
    "repliesTo": {
      "msg": "xhoAEBPnmEN2+kfTYrUrb6G0Q/Nc7x9lc0WB9UskoGk=.blake2s",
      "rel": "replies-to"
    },
    "mentions": [
      {
        "feed": "wuDDnMxVtk8U9hrueDj/T0itgp5HJZ4ZDEJodTyoMdg=.blake2s",
        "rel": "mentions",
        "name": "dominic"
      }
    ]
  }
```

changes to--

```json
  "content": {
    "type": "post",
    "text": "hello @dominic",
    "repliesTo": {
      "msg": "xhoAEBPnmEN2+kfTYrUrb6G0Q/Nc7x9lc0WB9UskoGk=.blake2s"
    },
    "mentions": [
      {
        "feed": "wuDDnMxVtk8U9hrueDj/T0itgp5HJZ4ZDEJodTyoMdg=.blake2s",
        "name": "dominic"
      }
    ]
  }
```

In this case, it is structurally similar -- it's only dropped the `rel`. Other messages will have to change structure. For instance:

```json
"content": {
  "type": "follow",
  "rel": "follows",
  "feed": "wuDDnMxVtk8U9hrueDj/T0itgp5HJZ4ZDEJodTyoMdg=.blake2s"
}
```

Will become

```json
"content": {
  "type": "follow",
  "follows": {
    "feed": "wuDDnMxVtk8U9hrueDj/T0itgp5HJZ4ZDEJodTyoMdg=.blake2s"
  }
}
```

--

As an aside, I'm also considering a semantic refactor for the message types, in which case follow may become something more like:

```json
"content": {
  "type": "follow",
  "newFriend": {
    "feed": "wuDDnMxVtk8U9hrueDj/T0itgp5HJZ4ZDEJodTyoMdg=.blake2s"
  }
}
[...]

PR #96, opened 2015-02-28
update to new link convention (bump ssb-msgs to 3.0.0, update tests)

pfraze:
https://github.com/ssbc/secure-scuttlebutt/issues/94

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,9 +136,29 @@ the key pair for this feed.
 
 ### SecureScuttlebutt#getPublicKey(id, cb)
 
-Retrive the public key for `id`, if it is in the database.
+Retrieve the public key for `id`, if it is in the database.
 If you have replicated id's data then you will have the public key,
 as public keys are contained in the first message.
+
+### SecureScuttlebutt#needsRebuild(cb)
+
+Checks the version stored in the database against the code version and
+calls back true/false accordingly. This keeps the database in sync with
+major breaking changes to secure-scuttlebutt when they occur.
+
+Should be run at startup. If true, you should call `rebuildIndex` before
+using the database.
+
+```js
+ssb.needsRebuild(function (err, b) {
+  if (b)
+    ssb.rebuildIndex(next)
+})
+```
+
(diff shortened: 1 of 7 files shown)


dominictarr:
hmm, you are gonna put a hyphen as the key? were we using replyTo on the old stuff?

pfraze:
as discussed in another thread, I used `repliesTo` instead of `replies-to`

the `indexLinks` function from `ssb-msgs` gives the reltype in all lowercase, which means ssb will lowercase its indexes. I'm going to update the index fetch functions to do the same on its inputs. this should have the effect of making the reltype case-insensitive

dominictarr:
Hmm, on second thoughts, this will actually create more complications elsewhere.
I think we are better off making things easy for ourselves at this point, rather than users.
If we did want to do something like this, we could enforce that keys are when added?
Or just allow case. What will this buy us?

pfraze:
@dominictarr ok have a look at the index rebuild code in the last commit there

pfraze:
@dominictarr updated `rebuildIndex`. I didnt go with the sync file because ssb isnt aware of the `~/.ssb` directory and it didnt seem very elegant to pass in a file path for storing metadata

dominictarr:
good call.

okay merging this.

dominictarr:
merged into 9.0.0

dominictarr:
oops, didn't have all the commits.
One thing here is you are pushing your code to a branch on the ssbc repo - there is nothing wrong with this perse, but the git commands to merge this are more annoying (fetch, merge) rather than if you use another repo (pull)

dominictarr:
this is fixed in 9.0.1

pfraze:
@dominictarr i can use forks if you like. I do merges through github which papers over that annoyance

also, make sure you push the merged repo to github, this is still showing as unmerged

dominictarr:
hmm, it's better to merge in the terminal so you can test. either way, using forks will still work with gh merge just the same.

PR #101, opened 2015-03-22
refactor createFeed out

dominictarr:
as discussed here: https://github.com/ssbc/ssb-keys/pull/12


dominictarr:
this is backwards compatible, btw.

pfraze:
ok looks good

dominictarr:
alright, merging.

dominictarr:
merged into secure-scuttlebutt@10.0.0 this was a breaking change since ssb-client depended on secure-scuttlebutt/feed which is now deleted. will make a pr to ssb client to update.

PR #103, opened 2015-03-23
update pull-level to get sync option

dominictarr:
update pull-level to get [{sync: true}](https://github.com/dominictarr/pull-level#example-realtime-aggregation)
which will enable solving this https://github.com/dominictarr/level-memview/issues/1 easily

### DIFF
--- a/index.js
+++ b/index.js
@@ -286,10 +286,10 @@ module.exports = function (db, opts) {
         lte:  [id, MAX_INT],
         live: live,
         keys: false,
-        sync: opts.sync
+        sync: opts && opts.sync
       }),
       paramap(function (key, cb) {
-        if(key.sync) return cb(key)
+        if(key.sync) return cb(null, key)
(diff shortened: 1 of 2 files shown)


dominictarr:
I started this and thought it would be easy, but then I realized I had created a race condition, because log stream does the lookup _after_ reading the live value.

This means that a write to a view cannot be up to date until some (async) time after the callback to the write has fired. Such behavior is not very useful for views, because we want views to be consistent.

The solution was to make a special case for `createLogStream` it basically does what pull-level [does](https://github.com/dominictarr/pull-level/blob/master/index.js#L47) but it uses the log index for the old part of the stream.

This means indexes can be rebuilt in the correct order, but still keeps everything very simple!

dominictarr:
oh looks like this is failing a test...

dominictarr:
okay this is fixed now.

dominictarr:
merged into 10.1.9

PR #107, opened 2015-04-28
Add list of other projects in the same space.

ootoovak:
I have always liked the community vibe when projects list other projects in the same space. It is a way of saying this idea is important/interesting enough that other people are also working on solutions. We like our solution but we acknowledge there are others out there doing great work also. 👍😀

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,3 +251,11 @@ it's decendant messages. If `parent` is true then each level will have
 ## License
 
 MIT
+
+## Other Projects
+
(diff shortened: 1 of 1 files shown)


dominictarr:
+1 also ipfs and firestar

ahdinosaur:
:+1:. also [fosp](https://github.com/maufl/fosp-specification).

ahdinosaur:
[slick](https://github.com/slick-io/slick)

ootoovak:
I believe this is listed elsewhere in SSB community documentation so closing here.

PR #109, opened 2015-05-01
Avoid thread locks

pfraze:
There have been reports of slow RPC and HTTP responses.
One issue, new messages have signatures and prev-hashes validated in the thread.
If there's a batch of new messages, they don't release the thread to the event loop until all sigs/hashes have been validated.

This PR has a test showing the issue (100 messages lock the thread for almost 2s on my laptop)
I'll add commits to release the thread after each validation, so that this issue can't occur.


pfraze:
Nope, this was wrong.
Turns out the validation pipeline is releasing the thread.
The delay I measured is from the multiple `feed.add` calls, which has to generate the prev hash and the message signature.
That's not an issue.

Learning is fun!
closing

PR #116, opened 2015-07-01
first steps toward end to end

dominictarr:
@pfraze okay checkout the test to see what i'm doing for end-to-end,
first we have to pass the keys into secure-scuttlebutt (as third argument)
this is so that we can decrypt messages that need to be indexed.

Then, the client uses ssb-keys to decrypt the messages they want to view.

Maybe we can make that a bit better than the second part, but at least
by explicitly decrypting we can see who knows what.


pfraze:
looking good

PR #117, opened 2015-07-03
User stream

dominictarr:
read a single user's stream in any range or order.

### DIFF
--- a/index.js
+++ b/index.js
@@ -262,6 +262,7 @@ module.exports = function (db, opts, keys) {
     var _values = opts.values
     opts.keys = false
     opts.values = true
+
     return pull(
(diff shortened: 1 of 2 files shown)


pfraze:
So why do we have a separate `createHistoryStream` from this? Couldnt we update `createHistoryStream` to work like this, and change the replicate plugin?

dominictarr:
Security. This is more flexible than a replication needs, so we'd want to limit the way that a replicator can use it, and it's simplest to do that via a separate method until we have arg permissions (if ever)

pfraze:
Ok makes sense, good to merge

pfraze:
published `11.2.0`

PR #120, opened 2015-07-08
Tracking: replace all the *Linked* functions with just one function.

dominictarr:
we talked about this a while back,
needed to use this again for blocking.


dominictarr:
merged into 13.0.0

PR #121, opened 2015-07-09
more througher end-to-end tests

dominictarr:
also added a comment because I couldn't see where the decryption was when i reread the code.


pfraze:
:cat2: :+1:

dominictarr:
okay, this is now a minor change. keys are passed through the ./create module, which is used by scuttlebot, and i've tested this working with scuttlebot.

dominictarr:
merging

dominictarr:
merged into 13.1.1

PR #123, opened 2015-07-16
fix edgecases when you know the source and the dest for a link

dominictarr:
sometimes you already know who may be linking to who, you just want to check whether they actually did.
this is needed in blocking in 2 instances (checking if A blocked B when B connects to A, or when B asks for A's feed)


pfraze:
:+1: looks good

dominictarr:
merge into 13.1.2

PR #125, opened 2015-07-24
Update links and link-indexing to use sigils

pfraze:
I ended up changing how the link indexing works in this. All links create the same 3 indexes, now:

 1. forward, from message author to link target
 2. backward, from link target to message author
 3. backward, from link target to message

This changes the `links()` function such that, if `type == msg`, then you'll get entries from index 3, and if `type == feed` then you'll get entries from index 1 and 2. `Type: msg` is now the default.

This change makes it possible to get links between feeds and messages/blobs.

However, this changes the meaning of the `type` parameter in `links()`. Instead of meaning, "give me links of that type," it's saying, "give me links *between* those types."

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,14 +14,14 @@ var pdotjson  = require('./package.json')
 var createFeed = require('ssb-feed')
 var cat       = require('pull-cat')
 var mynosql   = require('mynosql')
-var isRef     = require('ssb-ref')
+var ssbref    = require('ssb-ref')
 var ssbKeys   = require('ssb-keys')
 
 var Validator = require('ssb-feed/validator')
 
-var isFeedId = isRef.isFeedId
-var isHash = isRef.isHash
-//this makes msgpack a valid level codec.
+var isFeedId = ssbref.isFeedId
+var isMsgId  = ssbref.isMsgId
+var isBlobId = ssbref.isBlobId
 
 //var u         = require('./util')
 
@@ -127,50 +127,24 @@ module.exports = function (db, opts, keys) {
         value: id, type: 'put', prefix: indexDB
       })
 
-    mlib.indexLinks(content, function (link, rel) {
-      if(isFeedId(link.feed)) {
-        add({
-          key: ['feed', msg.author, rel, link.feed, msg.sequence, id],
-          value: link,
-          type: 'put', prefix: indexDB
-        })
-        add({
-          key: ['_feed', link.feed, rel, msg.author, msg.sequence, id],
-          value: link,
-          type: 'put', prefix: indexDB
-        })
-      }
-
-      if(isHash(link.msg)) {
-        // we don't need a forward index, because
-        // you can just parse the message easily.
-
-        // index from the linked message back to the
-        // linking message.
-        add({
-          key: ['_msg', link.msg, rel, id], value: link,
-          type: 'put', prefix: indexDB
-        })
-      }
-
-      //TODO, add ext links
-      //hmm, these should really 
-
-      if(isHash(link.ext)) {
-        //this is ACTUALLY ext from message!
-        //not ext from feed! TODO are there tests
-        //IS THIS USED?
-        add({ //feed to file.
-          key: ['ext', id, rel, link.ext], value: link,
-          type: 'put', prefix: indexDB
-        })
-        add({ //file from feed.
-          key: ['_ext', link.ext, rel, id], value: link,
-          type: 'put', prefix: indexDB
-        })
-
-      }
-
+    mlib.indexLinks(content, function (obj, rel) {
+      add({
+        key: ['feed', msg.author, rel, obj.link, msg.timestamp, id],
+        value: obj,
+        type: 'put', prefix: indexDB
+      })
+      add({
+        key: ['_feed', obj.link, rel, msg.author, msg.timestamp, id],
+        value: obj,
+        type: 'put', prefix: indexDB
+      })
+      // we don't need a forward msg index, because
+      // you can just parse the message easily.
+      add({
+        key: ['_msg', obj.link, rel, id, msg.timestamp],
+        value: obj,
+        type: 'put', prefix: indexDB
+      })
     })
   }
 
@@ -448,17 +422,19 @@ module.exports = function (db, opts, keys) {
     if(!opts.values&&!opts.meta&&!opts.keys)
       throw new Error('makes no sense to return stream without resultts'
         + 'set at least one of {keys, values, meta} to true')
-    if(!/^(?:msg|feed|ext)$/.test(opts.type))
-      throw new Error('must pass a type, feed|msg|ext')
-
+    if(!/^(?:msg|feed)$/.test(opts.type))
+      throw new Error('must pass a type, feed|msg')
 
     var type, rel, back
     var src = opts.source || null
     var dst = opts.dest || null
     var rel = opts.rel
-    var type = opts.type || 'feed'
+    var type = opts.type || 'msg'
     if(dst && !src) back = true
     var _type = back ? '_'+type : type
+
+    if (type == 'msg' && !back)
+      throw new Error('doesnt make sense to pass src for type: msg - just look in the message itself')
 
     return pull(
       pl.read(indexDB, {
--- a/index.js
+++ b/index.js
@@ -129,19 +129,12 @@ module.exports = function (db, opts, keys) {
 
     mlib.indexLinks(content, function (obj, rel) {
       add({
-        key: ['feed', msg.author, rel, obj.link, msg.timestamp, id],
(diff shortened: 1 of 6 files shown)


dominictarr:
okay this looks good.
hmm, i think we may not need the _msg index anymore
because you the _feed index is [link, rel, feed, seq, msg_id]
so you can query all the messages that link to a given object with that anyway

pfraze:
@dominictarr ok I need some insight on how ordering and range queries work

After your suggestion, here's what the index looks like:

```js
mlib.indexLinks(content, function (obj, rel) {
      add({
        key: ['link', msg.author, rel, obj.link, msg.timestamp, id],
        value: obj,
        type: 'put', prefix: indexDB
      })
      add({
        key: ['_link', obj.link, rel, msg.author, msg.timestamp, id],
        value: obj,
        type: 'put', prefix: indexDB
      })
})
```

So, here are the possible range queries:
```js
// 1
links({ type: 'feed', src:, rel: })
gte: ['link', src, rel, LO, LO, LO]
lte: ['link', src, rel, HI, HI, HI]

// 2
links({ type: 'msg', src:, rel: }) 
gte: ['link', LO, rel, LO, LO, src]
lte: ['link', HI, rel, HI, HI, src]

// 3
links({ type: 'feed', dst:, rel: }) 
gte: ['_link', dst, rel, LO, LO, LO]
lte: ['_link', dst, rel, HI, HI, HI]

// 4
links({ type: 'msg', dst:, rel: }) 
lte: ['_link', dst, rel, LO, LO, LO]
gte: ['_link', dst, rel, HI, HI, HI]

// 5
links({ type: 'feed', src:, dst:, rel: })
lte: ['link', src, rel, dst, LO, LO]
gte: ['link', src, rel, dst, HI, HI]

// 6
links({ type: 'msg', src:, dst:, rel: })
lte: ['link', src, rel, LO, LO, dst]
gte: ['link', src, rel, HI, HI, dst]
```

(Of course, `rel` is optional, so there are really 6 more variants without `rel`)

Queries 2 and 6 are not very important, because you can just as easily look in the source message itself for the links.
That said, would 2 and 6 work as intended? I suspect they'd fetch a lot of incorrect entries.

Regarding ordering, almost none of these will actually sort by `msg.timestamp`, right? Only if you provide `src`, `dst`, and `rel` would that happen. Otherwise, an earlier index in the key would take sorting precedence.

dominictarr:
to get 2 and 6 to work right you need to scan and fliter. obviously that is read efficient than having an index, but it's a trade off and 2,6 are queries that you have less often

pfraze:
@dominictarr Yeah. It would end up scanning nearly the entire index, which isn't very good at all. Maybe we should just give an error for 2 and 6, say "not supported, just look at the source message" ?

Also, am I right about the output order?

dominictarr:
hmm, i think this might not be quite right - one thing is the `type` works differently than it did before,
previously the type was required because it split the index, which was different. this should potentially be more simple and flexible.

I don't think we should error. it probably won't be much a problem.

does this pass the current tests? there is a fair amount of coverage for the link tests.

pfraze:
Tests do pass. Here's how type is working now (I believe!):

- `links({ type: msg, src: })` gives links from the `src` msg to their target feed/msg/blobs.
- `links({ type: msg, dst: })` gives backlinks from the `dst` feed/msg/blob to the linking message's ID.
- `links({ type: feed, src: })` gives links authored by `src` about other users.
- `links({ type: feed, dst: })` gives backlinks about `dst` authored by other users.

pfraze:
So, you use `type: feed` to get relationships between users

dominictarr:
@pfraze idea
```
ssb.links({source: &blobhash, dest: "@"}) //dest has sigil without full key, to indicate you want the feeds that linked to this blob.
```

pfraze:
@dominictarr ah that's a cool idea, but, I think that would serve a different function than `type: feed`.

That looks like it would run a filter on the outputs, so that you only get outputs of the sigil type...

...whereas, `type:feed` changes the index used, so that you can specify the author as the `src` or `dest`, rather than the message.

I think both would be useful

dominictarr:
ah, no you don't have to use the whole value as the key, so you could use 
``` js
{dest: '&' }
```
to get all messages that link to a blob.
that would become this range:
``` js
{
  gte: ['link', '&!', LO, LO, LO],
  lte: ['link', '&~', HI, HI, HI]
}
```
which isn't the same as filtering the output. of course, if you have a source and a rel that would work too.
When i made type:msg/feed it was intended to specify what type of link it was,
so, `dest: sigil` would have the same effect.

oops, just realized that the example i gave in the previous message was wrong, because a blob cannot be the source of a link.

pfraze:
@dominictarr yes definitely, I was just describing the end-effect.

> When i made type:msg/feed it was intended to specify what type of link it was, so, dest: sigil would have the same effect.

Right, I think this is a good idea. But, I think we should also have `type: feed`, because it lets you query by author. With `type: feed` you can say:

 1. Give me all links (optionally, of a certain `rel`) published by user X.
 2. Give me all links (optionally, of a certain `rel`) published by user X about user Y.

Neither of which we could do in `type: msg`, even with the sigil prefix.

dominictarr:
All links published by user alice (would include msgs and blobs)
``` js
ssb.links({source: @alice})
```
all links published from alice to bob
``` js
ssb.links({source: @alice, dest: @bob})
```
all links by alice to a message
``` js
ssb.links({source: @alice, dest: %})
```

dominictarr:
Oh, by the way... the index to be used is chosen by whether `source` or `dest` is not provided,
obviously, using the index that will give you the information that is missing.

pfraze:
Uh, wow, right. Duh.

Yeah, we can ditch type then.

dominictarr:
okay sweet!

pfraze:
@dominictarr I dont think we can reliably define output order, whether by author TS or local TS, or seq. Should we just remove that slot from the key?

dominictarr:
@pfraze the timestamp means that we can index multilpe links between alice -> bob in the same message. it could happen.

pfraze:
@dominictarr the timestamp wouldn't fix that, would it? We'd need a random number, or a counter that increments within the indexer for each link in the message.

dominictarr:
oh, sorry my bad. that isn't what the seq is for.
the seq means that the links between alice and bob are in a consistent order, it would still be disabmiguated by the message id, but that in uniformly distributed because it's a hash.
so the sequence changes it to be in the order of messages, per the sequence at least.

I don't think it will break anything to change it to timestamp...

dominictarr:
but im not sure it will be useful either

pfraze:
@dominictarr yeah, I'm not sure it's useful. The only time it would dominate order is for a query between a `src` feed and `dest` feed.

I suggest we do the counter, to support multiple links of the same rel and target in a message.

dominictarr:
@pfraze i think we could definately get multiple links between A and B.
say, A follows B, then unfollows B, then follows B again. the order is signifficant.

pfraze:
@dominictarr that's true, I just dislike that the query output won't be ordered the same way every time. For instance, a query with just `src` or `dst` would have an apparently random order, wouldn't it?

We could move the `seq` to to slot 1, right after the `link/_link` value, couldn't we?

dominictarr:
@pfraze it would be ordered by next time in the index. putting seq at the start would mean you'd have to scan past that which would destroy most of the purpose of the index. if you really need a particular order, for a wide query, sort it after you get the result.

Of course, with such a wide query, order will not be well defined anyway, since ssb messages arn't well ordered between feeds. However, having seq/ts at least means that the indexes are ordered in the cases where ssb _is_ well ordered, which feels reasonable.

pfraze:
@dominictarr Ah, yes, the range queries are tricky to reason about.

Ok, we should just be clear in the docs that the order is only defined for queries with `src`, `rel`, and `dest`, where `src` is a feed.

dominictarr:
yeah.

pfraze:
@dominictarr I think I implemented this the way you suggested, but the index range was pull out more hits than needed

A post-fetch filter enforces the correct results https://github.com/ssbc/secure-scuttlebutt/blob/link-sigils/index.js#L499-L507. I left that in, with some debugging logs

If you comment out the post-filter and run the link-types test, you'll see the error

If I didn't mess up the implementation, then the issue is, without a `rel` value, the type-based gte/le values dont work

dominictarr:
yeah, this looks correct. there will be filtering required in some queries (like if you leave off rel)

(we could add an index for rel, but we'd need 6 indexes, `{src,rel,dst}, {src,dst,rel}, {dst,rel,src}, {dst,src,rel}, {rel,src,dst}, {rel,dst,src}` instead of just two indexes.

dominictarr:
merged into 14.0.0

PR #127, opened 2015-09-15
update createFeedStream range params to operate on timestaps

pfraze:
I noticed that the gt/gte/lt/lte params were operating on the index without modification. Well, the feed index looks like `[ts, author]` so there's not much reason to query against that. This PR makes those params query against just `ts`

### DIFF
--- a/index.js
+++ b/index.js
@@ -221,6 +221,9 @@ module.exports = function (db, opts, keys) {
   //TODO: eventually, this should filter out authors you do not follow.
   db.createFeedStream = function (opts) {
     opts = stdopts(opts)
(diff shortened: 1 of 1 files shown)


dominictarr:
would this be better with indexing on the local timestamp? because then it would be easy to sync and see when you have new data. you could still choose to insert it at an older place, if necessary. oh no, that would just make it the same as logdb... hmm, would it be better to just use log stream for this?

dominictarr:
merged into 14.1.0

pfraze:
Yeah this index gets kinda scrwed up as soon as somebody puts a bad timestamp in there (eg 0 or `MAX_INT`)

Issue #129, opened 2015-09-23
getPublicKey is incorrect


pfraze:
`getPublicKey` is no longer correct, because we don't publish an `init` message anymore.

We can either rewrite it to extract the pubkey out of the ID, or we can remove it. I'm in favor of the former, in case we decide to use pubkey hashes again in the future.

dominictarr:
lets just remove it. We'd have to commit to that possibility and tell everyone to call that method, but they wouldn't because it's easier to just grab it off the message, indeed that is what we do internally!

stability levels are described in the stability module, which i pulled from the node.js stability index.
which seems quite reasonable, after a discussion about nuances of semver (whether v1.0 has a special meaning)

[unstable](https://github.com/dominictarr/stability#unstable) means there may be features, patches, and possibly breaks, [stable](https://github.com/dominictarr/stability#stable) means do not expect breaks.

I'm inclined to say `stable`, given the choices, but in all honesty, I do think we may have a breaking change. If it was entirely up to me, i'd lable it "pretty stable". Although, that is not part of the stability index spec!

HOWEVER! I am the maintainer of the stability module, and it's stability is marked as _stable_ so we could make _additions_ to the spec without contradicting the meaning of "stable", and we can add various qualifiers for the manner of the maintainers eagerness to change the module, then we could call it "pretty stable" or "hopefully stable". We can't rule out breaking changes, but hopefully won't make them.

pfraze:
:+1:

PR #130, opened 2015-09-23
update readme

pfraze:
Mostly typo fixes and some modifications for clarity. Major changes:

 - Mark stable. I believe this is the case? We're definitely out of experimental, I think, but maybe we should be unstable, not stable. (https://github.com/dominictarr/stability#levels-of-stability)
 - Update the API to show the new `links` API, instead of the old split up link-queries.

Still to decide:

 - Are we stable or unstable?
 - Should we remove getPublicKey? #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,52 +2,65 @@
 
 A secure database with replication that is guaranteed to work.
 
-## stability: level 1, experimental - expect breaking changes.
+## Stability
 
-Following the [node.js stability index](https://github.com/dominictarr/stability#levels-of-stability),
-and the good parts of [semver](http://semver-ftw.org)
-v1 does not mean stability, it just means there has been a breaking change
-since v0.
+Stable: Expect patches, possible features additions.
 
 ### Documentation/wiki/FAQ
 
-[documentation is here](https://github.com/ssbc/ssb-docs)
+**[Documentation is here](https://github.com/ssbc/ssb-docs)**.
+
 We have shifted documentation from a github wiki to a repo,
 which means you can ask make pull requests, get notifications,
 ask questions in issues. If you have questions or get confused
 please post an issue!
 
-## example
+## Example
 
 
 ``` js
-// create a scuttlebutt instance and add a message to it.
+/**
+ * create a secure scuttlebutt instance and add a message to it.
+ */
 
 var pull = require('pull-stream')
+
+// paths:
+var pathToDB     = '/tmp/ssb1/'
+var pathToSecret = '/tmp/ssb1-secret'
+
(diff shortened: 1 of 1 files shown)


dominictarr:
(accidentially posted this in the getPublicKey issue)

stability levels are described in the stability module, which i pulled from the node.js stability index.
which seems quite reasonable, after a discussion about nuances of semver (whether v1.0 has a special meaning)

[unstable](https://github.com/dominictarr/stability#unstable) means there may be features, patches, and possibly breaks, [stable](https://github.com/dominictarr/stability#stable) means do not expect breaks.

I'm inclined to say `stable`, given the choices, but in all honesty, I do think we may have a breaking change. If it was entirely up to me, i'd lable it "pretty stable". Although, that is not part of the stability index spec!

HOWEVER! I am the maintainer of the stability module, and it's stability is marked as _stable_ so we could make _additions_ to the spec without contradicting the meaning of "stable", and we can add various qualifiers for the manner of the maintainers eagerness to change the module, then we could call it "pretty stable" or "hopefully stable". We can't rule out breaking changes, but hopefully won't make them.

dominictarr:
that said, we have already broken this a lot of times. so I think that speaks for itself. although, it is worth pointing out that there is a lot of dependent code, so breaking this module is not something we do on a whim.

dominictarr:
I have merged this, and removed getPublicKey since it wasn't actually used at all, so the corresponding breaking change has already been recorded. removing a feature that was already broken and no one is using is not a breaking change.

dominictarr:
merged into 14.1.1

ahdinosaur:
ooh, `stable`, that's a statement, does that mean as an app developer i can use `secure-scuttlebutt` without worrying about being pained by a breaking change? ;) i'd lean more towards `unstable`, but if y'all are serious about backwards compatibility (at the very least through making migration tools as required by any breaking change), `stable` sounds great to me.

pfraze:
Yeah @ahdinosaur that's definitely the question. I don't think it's quite right to say we'll never make breaking changes, because that's like saying we'll never bump major. But, I think the fundamentals of the protocol are stable enough at this point, that, big changes would be backwards compat with the old *data*. That sound right @dominictarr ?

dominictarr:
@pfraze yeah, once we make a big public announcement about it, we are definately commited to mostly backwards compatible data (migrations, fallbacks, etc). But this is still new, so we may want to change some apis or something.

I think what we could do, is have backwards compatible data, but only allow the old version with also a low enough timestamp.

which is why i'd say "pretty stable" or "probably stable"

PR #131, opened 2015-10-21
rewrite links

dominictarr:
After noticing that links was not handling some cases the other day, (in particular, queries where not filtered by rel at all) I rewrote links.

This bug was introduced by upgrading links to use sigils.
https://github.com/ssbc/secure-scuttlebutt/pull/125

This makes a breaking change: if a `source` (or source sigil) was not provided,
previously it defaults to the message as the source. but the author as the source makes more sense. You can get the message easily - from the `.key`
property (and have links return you the whole message with values: true)
Making source always be a feed is also more consistent with the query
input. It doesn't really make sense to query the source as a message
anyway.

This did mean changing the tests, but I grepped through patchwork and scuttlebot and we havn't actually used a source value in a links query (not in the product code, there are examples in the tests)

there are some times when you might want to do so - like, to check if Alice has blocked Bob, you could query `links({source: alice, dest: bob, rel: follows})` but in the cases we are doing that we have an in memory structure, so this change wouldn't actually break any code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -440,15 +440,17 @@ module.exports = function (db, opts, keys) {
           dest: op.key[back?1:3],
           key: op.key[5]
         }
-      }),      // apply any filters
+      }),
+      // in case source and dest are known but not rel,
+      // this will scan all links from the source
+      // and filter out those to the dest. not efficient
+      // but probably a rare query.
       pull.filter(function (data) {
         if(rel && rel !== data.rel) return false
         if(!testLink(data.dest, dst)) return false
         if(!testLink(data.source, src)) return false
         return true
       }),
-      //handle case where source and dest are known but not rel.
-      //this will scan all links from the source. not so efficient.
       ! opts.values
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -397,105 +397,58 @@ module.exports = function (db, opts, keys) {
...
-      (srcfilter || dstfilter) ?
-        pull.filter(function (d) {
-          if (srcfilter && ssbref.type(d.source) != srcfilter)
-            return false
-          if (dstfilter && ssbref.type(d.dest) != dstfilter)
-            return false
-          return true
-        }) : null,
       //handle case where source and dest are known but not rel.
       //this will scan all links from the source. not so efficient.
```
pfraze: we should remove this comment if the code is going to be removed

pfraze:
much clearer! :+1:

index.js
dominictarr: fixed

PR #132, opened 2015-10-29
apis needed for replication back pressure

dominictarr:
This changes a few apis as needed for replication back pressure.
This is a breaking change, but only changes a few small features.

* `latest` now returns tuples of `{sequence, ts}` instead of `sequence`. data created prior to this merge won't have ts, so it's left as `null` and replication can just handle that by allowing the maximum number of messages, and this will be updated as soon as one is received. `ts` is the local received time.

* `getLatest` is removed. This wasn't being used anymore.

and some feature additions

* `createHistoryStream` now accepts an integer `limit` property. incidentially, this means we can later add partial ranges and it won't be a breaking change.

Also, I added test coverage for `db.latest` and removed some dependencies which are no longer used.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,9 @@ var isBlobId = ssbref.isBlobId
 //53 bit integer
 var MAX_INT  = 0x1fffffffffffff
 
+function isNumber (n) {
+  return typeof n === 'number'
+}
 
 function isString (s) {
   return 'string' === typeof s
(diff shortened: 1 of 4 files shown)


dominictarr:
@pfraze rfr

pfraze:
:+1:

dominictarr:
merged into 15.0.0

PR #135, opened 2015-12-23
added ```#``` sigil to *type* check expression; a one-liner

NHQ:


pfraze:
I talked with @NHQ about this in IRC. The goal is to create a new index. I suggested creating the index in his application code, and we talked about a few possibilities to make it more convenient. I'm actively exploring options for this now, with Patchwork.

PR #136, opened 2016-01-04
Edit README to make it more beginner-friendly

paulkernfeld:
I think that almost all of this information is correct, but please let me know if anything in here is inaccurate or misleading. In particular, I didn't quite understand how encryption and SSB interact.


README.md
```diff
@@ -80,39 +81,88 @@ pull(
...
+### Messages
 
-### Secure Data Structures
+Each message contains:
 
-SecureScuttlebutt uses a signed block-chain per identity.
-Each block points to the previous block,
-the signing key, and contains a short message
-and a signature. Every identity has their own block-chain.
+- A message string. This is the thing that the end user cares about.
```
pfraze: Unless encrypted, it's a message object.

README.md
```diff
@@ -80,39 +81,88 @@ pull(
...
+- A reference to the previous message. This prevents a malicious party
+  from making a copy of a feed that omits messages in the middle.
```
pfraze: Close but that's not the actual reason for the prevhash. The signature + sequence numbers are enough to stop middlemen from dropping messages. The reason for the prevhash is (and my suggested copy):

```
A content-hash of the previous message. This prevents somebody with the private key from changing
the feed history after publishing, as newly-created message wouldn't match the "prev-hash" of latter messages.
```

pfraze:
Overall great updates, IMO. Just needs the fixes I suggested. @dominictarr thoughts?

README.md
paulkernfeld: Ah, got it. Thanks!

README.md
paulkernfeld: Oops! I'll put that in.

pfraze:
@dominictarr I suggest we merge

paulkernfeld:
Bumping: should this be merged, edited, or closed?

dominictarr:
sorry I missed this, reviewing now...

pfraze:
I used this as one of my references when I put together http://ssbc.github.io/secure-scuttlebutt/, which now lives in this repo's `gh-pages` branch as https://github.com/ssbc/secure-scuttlebutt/blob/gh-pages/index.md

dominictarr:
this looks great.

two small niggles. in links, dest can be pubkey or hash (or feedId, msgId, blobId) and source can be feedId.

in the intro you call it a kappa architecture. I kinda hate this term because "kappa" is just an arbitary greek letter so it doesn't carry any descriptive power. For things that have lots of arbitaryness a arbitary name is good, take for example "rugby". rugby is a sport where you run around with a ball, and the name doesn't mean anything. On the other hand "football" is a descriptive name (except in american football where you mostly carry the ball). In that case, I think football is a bad name, because there are a lot of things you could change and it wouldn't be soccer anymore, but it would still involve feet and balls. On the other hand, most track and field sports are really simple and suit descriptive names (100 meter sprint) though occasionally non-descriptive names sneak in (marathon)

Also, pretty much every database is "kappa architecture" (at least on the inside) all it needs to qualify for that is to expose a log of changes. We are currently discussing what sort of database secure-scuttlebutt is here in the [indexes issue](https://github.com/ssbc/secure-scuttlebutt/issues/142)

dominictarr:
But I'm gonna merge this and then change those bits, that is just my opinion about that name and this is otherwise great work!

dominictarr:
oh yeah, and relatedMessages takes a msgId not a feedId

dominictarr:
merged into 15.0.2

PR #143, opened 2016-01-28
change entities to objects

wanderer:
"entities" in this sentence confused me. Because an entity "is something that exists in itself". But feeds are composed of messages so they they are a single entity. "Objects" on the other hand is a more general term.

### DIFF
--- a/index.md
+++ b/index.md
@@ -61,7 +61,7 @@ Here's an example message:
 
 ### Entity References (Links)
 
-Messages can reference three types of Secure Scuttlebutt entities: messages, feeds, and blobs (i.e. files).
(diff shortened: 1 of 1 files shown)


pfraze:
I'm :+1: this, we usually use the term object

dominictarr:
yup

dominictarr:
@pfraze if you are gonna merge can you publish a new version? if it's only a small documentation change that is a patch.

pfraze:
yeah, this got merged into `gh-pages`, not master, which is a mistake - we should treat `master` as canonical and merge changes into `gh-pages` after. I'll fix that and publish the update

PR #144, opened 2016-02-22
get accepts a timestamp

dominictarr:
in the streamview work, everything is based around an append only log, which provides order, but is also the main storage of the database. Since it's append only storage, it makes sense to be able to use a monotonically increasing key. We can make this more efficient later, by putting the data and the log into an append only file, but this is enough to allow development with streamviews.

It should be stressed though, that this is only intended for internal apis to use. the timestamp you have will not be the same as someone elses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,13 +57,20 @@ function getVMajor () {
 
 module.exports = function (db, opts, keys) {
   db = mynosql(db)
+  //small data
   var sysDB   = db.sublevel('sys')
+  //append only
   var logDB   = db.sublevel('log')
-  var feedDB  = db.sublevel('fd')
+  //sorted
+  var feedDB  = db.sublevel('fd') //NOT USED.
+  //sorted: [author, seq] => id //immutable
   var clockDB = db.sublevel('clk')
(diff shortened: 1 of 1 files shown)


pfraze:
Cool, what motivated adding this? I assumed get(hash-id) would be enough

Also would you mind commenting around each of the indexes (in the code) what each of them do and why some are deprecated?

dominictarr:
I want to try this because it would allow us to adopt an architecture where the main log was just literally, one big log. everything would go in one file, and always be appended to. there would be other views on the database but they are generated from the log. that would mean there is just one file to back up, so it's really easy to back up. if I want to [sneakernet](https://en.wikipedia.org/wiki/Sneakernet) or [pidgeon post](https://en.wikipedia.org/wiki/Pigeon_post) you can just do that one file.

In this case, it's more efficient to use the ordered, append-only, key as the primary key, because then you can do the lookup in one step.

Infact, if the key is based on the offset in the log + the size of the object, then you can have reads from that primary store that are `O(1)` instead of `O(log(n))`!

So, I want to try this basically to have a chance to experiment towards this architecture.

PR #146, opened 2016-03-31
Ignore missing root message in relatedMessages()

pfraze:
Prior to this PR, if you ran `relatedMessages()` on a msg-id that's not present†, but which does have present messages linking to the root, then `relatedMessages()` would call the cb() with an error message, and, in some cases, would call the cb *twice*, resulting in a backend crash (cb called twice error).

This PR does two things:
 - Guards against the double cb-call
 - Ignores a root-not-found error, so that the rest of the thread can be assembled

† present: saved in the local database

### DIFF
--- a/index.js
+++ b/index.js
@@ -483,12 +483,15 @@ module.exports = function (db, opts, keys) {
     var msgs = {key: key, value: null}
     db.get(key, function (err, msg) {
       msgs.value = msg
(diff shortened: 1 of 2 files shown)


clehner:
:clap:

dominictarr:
awesome, merged into 15.0.5

PR #147, opened 2016-04-14
remove mynosql

dominictarr:
mynosql was a cute idea, but streamviews are much better.
we never actually used mynosql features, so it's easy to remove it.
I think we can consider this a patch.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,11 +9,9 @@ var assert    = require('assert')
 var ltgt      = require('ltgt')
 var mlib      = require('ssb-msgs')
 var explain   = require('explain-error')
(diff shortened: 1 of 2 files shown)


dominictarr:
this was causing problems because the indexes and the log did not have exactly the same timestamp.
this won't fix the built in indexes for links, but it won't be a problem for streamview based things, since they rebuild from the log. So, this is working towards the point to depreciate the links api, and just use indexes as plugins.

pfraze:
👍

ahdinosaur:
> mynosql was a cute idea, but streamviews are much better.

:sparkles:

PR #149, opened 2016-05-20
towards consistent behaviour for live streams

dominictarr:
I want all of the streams that read stored data + changes to work consistently.
I have created a module https://github.com/pull-stream/pull-live that handles the edge cases
(so the stream can be read in old only, live only, or combined modes)

So far, I have only updated `createLogStream` because I agree with https://github.com/ssbc/patchwork-threads/issues/11#issuecomment-218302198 we should remove the indexes from secure-scuttlebutt and move them into plugins.


dominictarr:
(but createLogStream will always be there because that is what the plugins connect to)

dominictarr:
okay, got this working for links too.

dominictarr:
actually, it won't be too hard to fix for everything once I figured out how to do it for links

dominictarr:
I'm gonna merge this

dominictarr:
merged into 15.2.0

PR #152, opened 2016-07-05
msg

clehner:


dominictarr:
what does this fix?

`_value` is used so that live streams can emit immediately without a lookup.
https://github.com/ssbc/secure-scuttlebutt/blob/dc1c1f8a2000ba622881cb68de049541529eebe0/index.js#L479-L480

Which prevents a race condition in other places... though, I think that was before `pull-live` which is a better approach in general, so maybe that is now obsolete?

clehner:
It is to fix `NotFoundError` being emitted from `ssb.links`, discussed in [%rBu56IqW1RopjM7yOQUI3ckPjVC36b14ax8rLei2jVM=.sha256](https://git-ssb.celehner.com/%25rBu56IqW1RopjM7yOQUI3ckPjVC36b14ax8rLei2jVM%3D.sha256)

`_value` isn't actually being used until this line is added:
https://github.com/ssbc/secure-scuttlebutt/commit/dc1c1f8a2000ba622881cb68de049541529eebe0#diff-168726dbe96b3ce427e7fedce31bb0bcR457

I see `db.links` uses `pull-live` but I don't see how that should fix the race. I think what is happening is that before a message is added to the db, the pre hook function ([L74](https://github.com/ssbc/secure-scuttlebutt/blob/dc1c1f8a2000ba622881cb68de049541529eebe0/index.js#L74)) runs, passes a link index op synchronously through `realtime(kv)` ([L82](https://github.com/ssbc/secure-scuttlebutt/blob/dc1c1f8a2000ba622881cb68de049541529eebe0/index.js#L82)) to the pipeline in `db.links` ([L499](https://github.com/ssbc/secure-scuttlebutt/blob/dc1c1f8a2000ba622881cb68de049541529eebe0/index.js#L499)), to `lookupLinks` which calls `db.get(op.key, ...)` ([L481](https://github.com/ssbc/secure-scuttlebutt/blob/dc1c1f8a2000ba622881cb68de049541529eebe0/index.js#L481)). Trying to `get` the new message fails because the pre hook function hasn't returned yet and so the message hasn't yet been added.

dominictarr:
now i remember. at first, i implemented real time views by just creating a live stream of the index,
then mapping that through the lookup. but this meant that realtime view was always delayed (because it would do a `get`.

but the new code just does the realtime part by filtering `db.post(..)` so this should now be unnecessary.

merged into 15.2.4

clehner:
I just realized that the pull request message that I wrote for this wasn't published, because I used a wrong command-line flag (`-m msg` instead of `-F msg`) to the [hub](https://github.com/github/hub) command. Sorry about that. Anyway, if the errors I was seeing persist, I will revisit this and maybe try to write a test case.

dominictarr:
ah got it

PR #153, opened 2016-07-10
Fix realtime links

clehner:
Turns out that #152 didn't fix what it was supposed to. The changed code wasn't covered by tests, so I didn't notice that the fix was not correct. The code path that uses `_value` is activated using `ssb.links({values: true, live: true, ...})`. Here I've corrected the fix and added a test to cover it.


clehner:
@dominictarr is this ok to merge?

dominictarr:
sure. merged into 15.2.5

PR #155, opened 2016-10-30
Add missing dependency to package.json file.

OlegGirko:
This change adds NPM package that is used but not declared as dependency
to `dependencies` section of `package.json` file.

Patchwork fails to start without this change.


dominictarr:
thanks, this is correct. merged into secure-scuttlebutt@15.5.1

PR #158, opened 2017-03-21
Replace await with wait

staltz:
`await` is a reserved word, and is disallowed by recent JavaScript parsers, such as Babylon. This has nothing to do with strict mode.


dominictarr:
hmm, this is technically a breaking change since .await is an exposed property, however using grep, it doesn't seem like anything actually uses that. I also linked branch in as sbot's dep and ran sbot's test and it passed (which I normally do when changing secure-scuttlebutt) so I am gonna merge this.

PR #167, opened 2017-05-20
merge flume stuff

dominictarr:
this will bring in all the changes from the flume branch.
this is a small breaking change (rebuild is removed) and the api for plugins that create database views is different (use _flumeUse instead of handling it your self)
I'll make PRs to those other things too.


dominictarr:
So, if lots of people are gonna run this, I think I better write a thing that makes sure nothing bad happens while it's migrating from level to flume. I'm gonna do that, then this is fully ready!

dominictarr:
Okay, now when loading a sbot instance that already has a level database, this will not allow any reads or writes until the new log is filled in, then all views will be regenerated. It will take a while to reload, actually.

dominictarr:
okay, merged into 16.2.0

PR #172, opened 2017-06-12
Add missing dependency ltgt (used in indexes/last.js)

alanshaw:
Tests fail on Node.js 4


dominictarr:
wow, that is a really old node, but good catch anyway!

dominictarr:
merged into 16.3.3

PR #173, opened 2017-06-17
migration: switch to count based progress once total known

mmckegg:
This PR attempts to count the total number of posts, and then once the count is complete, switches progress mode from **timestamp** to **total number**. Turns out it is much quicker to stream through the entire database if you aren't doing anything besides counting. The following code returns in under 10 seconds on my computer:

```js
  var count = 0
  var now = Date.now()
  pull(
    context.sbot.createLogStream(),
    pull.drain(x => {
      count += 1
    }, () => {
      console.log(count, (Date.now() - now) / 1000) // => 142596 9.368
    })
  )
```

That's at least 15 times quicker than the actual migration. It means that within 10 seconds a newly onboarded user will start seeing actual progress instead of more like 1.5 minutes (which is how long @mischa-saunders had to wait before the progress started to tick up on his machine).

---

I can understand why this would be controversial. It is yet another one of my "slow things down for better progress" kind of PRs 😆 

Please discuss!! @ahdinosaur @dominictarr

### DIFF
--- a/legacy.js
+++ b/legacy.js
@@ -66,6 +66,8 @@ module.exports = function (db, flumedb) {
 
   if(flumedb) {
     var prog = {}
+    var prog2 = {current: 0, start: 0, target: 0}
+
     function one (opts, cb) {
       pull(
         db.createLogStream(opts),
@@ -78,6 +80,7 @@ module.exports = function (db, flumedb) {
     function update (since) {
       var start = (prog.start = prog.start ? prog.start : +since)
       prog.current = +since
+      prog2.current += 1
     }
 
     one({reverse: true, limit: 1}, function (err, last) {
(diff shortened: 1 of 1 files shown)


dominictarr:
we can probably make this even faster... this is still reading the index, then retriving each record and then throwing it away. we could just stream the index without doing lookups (and also skipping parsing) experimenting now, looks 10x faster

dominictarr:
okay, on test/migation.js this is 10x faster. If you set the test to create 100k messages https://github.com/ssbc/secure-scuttlebutt/blob/master/test/migration.js#L23 it counts in one second. I think this is fine. I made it wait until it has counted everything before starting the migration... because it was easier to see what was going on, and also it meant we can delete the other progress system, so less code to maintain!

mmckegg:
Ah nice!

dominictarr:
merged into 16.3.4

PR #174, opened 2017-06-26
throw immediately when `db.get` called with a non-ref string

mmckegg:
Right now if you call `get('someString')` things get really nasty. The request makes it all the way to https://github.com/flumedb/aligned-block-file/blob/master/blocks.js#L40 before throwing. It's also inside async so the stack trace is lost and you have no idea where the culprit is! (I wasted an hr trying to track down something thanks to this)

**This PR adds a check to ensure that the key is a number before passing it to flume. It throws an error if unhandled.**

_Warning: Code untested! Did this instead of an issue so the fix could be illustrated clearly. Not sure how to word the error message._

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,7 +89,10 @@ module.exports = function (_db, opts, keys, path) {
         if(err) cb(err)
         else cb(null, seq && seq.value)
       })
-    else _get(key, cb) //seq
+    else if(typeof key === 'number') 
+      _get(key, cb) //seq
+    else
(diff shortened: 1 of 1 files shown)


dominictarr:
since this is a programmer error, I would throw instead of cb the error. cb's are for io errors. anyway, will merge.

dominictarr:
when making small changes like this, I'd usually just run the tests, then symlink it into scuttlebot and run those tests too.

merged (but throws instead) into 16.3.5

PR #175, opened 2017-06-30
bump flumedb version

mmckegg:
to bring in the fixes from flumedb/flumedb#7

(this is pre 1.0 so even with the `^`, it is fixed at 0.2.x)

@dominictarr


dominictarr:
merged into 16.3.6

PR #176, opened 2017-06-30
fix resume migration progress

mmckegg:
If you resume from a migration (after quitting sbot midway), progress display restarts at 0, but work resumes from last stop. This means that the progress display appears to stall midway, but it is actually because it's comparing against the total number, not the remaining.

**There is no way to get the remaining number using the fast track count method, so instead this PR starts the migration from `gt: 0` and then skips all of the items before `since`. Now migration appears to resume from where it left off after a fairly quick "catchup"**

I think this small catchup time is worth the trade-off because resuming from migration is not that common and the UX of a stuck progress bar is basically the worst possible thing. 

cc @dominictarr

### DIFF
--- a/legacy.js
+++ b/legacy.js
@@ -111,10 +111,14 @@ module.exports = function (db, flumedb) {
         function migrate () {
           // actual upgrade
           pull(
-            db.createLogStream({gt: since}),
+            db.createLogStream(),
             paramap(function (data, cb) {
               prog.current += 1
-              flumedb.append(data, cb)
+              if (data.timestamp > since) {
+                flumedb.append(data, cb)
+              } else {
+                cb() // skip this one, we already have it!
+              }
             }, 32),
             pull.drain(null, ready)
           )


dominictarr:
hmm, I don't understand why it doesn't recalculate the number of items correctly? is it because it writes new receive time timestamps?

mmckegg:
@dominictarr It is because when you resume a migration there are less items to convert (`gt: since`), but it is [counting **all** of the items](https://github.com/ssbc/secure-scuttlebutt/blob/0ad3e7f86af08d530ad5aad6d6842d4ad918685c/legacy.js#L102) from the original db, and there is no way to filter the count by timestamp since we're not loading values.

dominictarr:
oh, yes there is because logDB is `timestamp -> hash` and we are loading the keys.
making another pr

mmckegg:
closing, #178 fixes this

PR #177, opened 2017-07-01
filter private messages out of into the log stream

dominictarr:
this simple change makes patchbay@6/minbay approach to private messages work much better. with out this, it loads 2.7 mb of messages on load, with this, only .5

### DIFF
--- a/index.js
+++ b/index.js
@@ -184,23 +184,26 @@ module.exports = function (_db, opts, keys, path) {
 
 
   db.createLogStream = function (opts) {
-    var limit = opts.limit
-    if(limit)
-      delete opts.limit
+    if(opts && opts.raw)
+      return db.stream(opts)
+
     opts = stdopts(opts)
-    if(opts.raw)
-      return db.stream()
+    var limit
+    if('boolean' === typeof opts.private) {
+      limit = opts && opts.limit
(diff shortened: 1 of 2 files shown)


dominictarr:
oops, realized I had npm published but hadn't pushed to master. just pushed to 04f7023

index.js
```diff
@@ -182,12 +185,23 @@ module.exports = function (_db, opts, keys, path) {
 
   db.createLogStream = function (opts) {
     opts = stdopts(opts)
+    console.log('createLogStream', opts)
```
mmckegg: console.log!

dominictarr: haha, sorry, fixed!

evbogue:
I'm not sure how I managed to tag this with 'bug' when I was looking at it, but now it's untagged.

dominictarr:
This is a good example of how NOT to write a pull request. The jerk (I can say that because it was me!) that made this PR implicitly expected that you'd read the code and understand the intention. It would have been much simpler to say "a 'private' option is added, and if it's set to true, only private messages are returned, but if it's false, no private messages are returned". 

"this simple change" does _not_ explain. I had forgotten about this and reading it again I couldn't figure out what this actually does. TISK TISK TISK past @dominictarr. closing

PR #178, opened 2017-07-01
oops, pass in the timestamp to the legacy migration

dominictarr:
@mmckegg this should do it

### DIFF
--- a/legacy.js
+++ b/legacy.js
@@ -99,7 +99,7 @@ module.exports = function (db, flumedb) {
         flumedb.progress.migration = prog
         var c = 0
         pull(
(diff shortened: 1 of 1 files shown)


mmckegg:
Oh, for some reason I thought old db was keyed by sequence? Is it timestamps? Awesome 🎉 

Guess I should have just tried it 😆

dominictarr:
the old db is keyed by hash, but this is the _logDb_ but maybe it's confusing because that line is `pl.old(logDB, {...})`

mmckegg:
Oh yes, of course 👍

dominictarr:
merged into 16.3.7

PR #186, opened 2017-10-29
Fix `isFunction`

### LINKED ISSUE #185: `isFunction` method incorrectly matching for type `funciton`
The `isFunction` function on line 34 in `index.js` is looking for type `funciton` rather than `function

emyarod:
The `isFunction` function on line 34 in `index.js` is looking for type `funciton` rather than `function. This pull request fixes that error and correctly matches against type `function`. Closes #185

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ function isString (s) {
 }
 
 function isFunction (f) {
-  return 'funciton' == typeof f
+  return 'function' == typeof f
 }
 
 var isArray = Array.isArray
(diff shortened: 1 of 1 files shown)


arj03:
I don't think the function is used. Which is probably why no-one has noticed this bug. Also the isNumber function can be removed as well. Care to update your PR @emyarod ?

emyarod:
Done @arj03 !

arj03:
Thanks! I think this looks good, but would like to have someone second this.

staltz:
Yes, looks okay to merge

arj03:
Thanks @emyarod!

PR #190, opened 2017-12-19
Modify createLogStream in index.js to be createRawLogStream

mlegore:
The createLogStream method in index.js is getting overwritten in extras.js, but I would like raw access to the method still, with range control, so I've renamed this one to a new method, so it is accessible.

### DIFF
--- a/index.js
+++ b/index.js
@@ -94,7 +94,7 @@ module.exports = function (_db, opts, keys, path) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


dominictarr:
yes this is a good idea. I want to clients to move towards using the raw stream because it will mean that we can remove a leveldb. merging.

dominictarr:
merged into 17.1.0

PR #199, opened 2018-04-06
pass decrypted messages to indexes

dominictarr:
This change decrypts all messages as they are read from the database.
Most current indexes will now work with private messages, so this means that you can privately follow or privately block some one, or privately name someone, or privately like something. All message types can now be private, and should pretty much just work (or maybe just small changes)

This is a big internal change, but I carefully went through and made each api that returns messages _not_ return decrypted messages by default. It would be okay if local clients such as patchwork got decrypted messages, but it could be a leak for web viewers.

All apis now have a `private` boolean option, which defaults to false and if set explicitly to true, decrypted messages are returned. decrypted messages have some extra properties:

a normal message looks like this:

``` js
{
  previous: msg_id,
  sequence: sequence,
  author: feed_id,
  timestamp: timestamp,
  content: ENCRYPTED_STRING
  signature: sig
}
```

``` js
{
  previous: msg_id,
  sequence: sequence,
  author: feed_id,
  timestamp: timestamp,
  content: {plain text object...}
  signature: sig,
  cyphertext: ENCRYPTED_STRING,
  private: true
}
```

if "private" is false, the cyphertext property is copied back to content, and the other extra fields removed. Note that this means a decrypted message is not a valid message, the signature will not verify and the hash is wrong, so in the worst case if it was sent over the network at least it would probably be dropped as invalid. However, all care is exercised to ensure that decrypted messages are not sent.


staltz:
Fantastic idea and PR! My excitement for it doesn't mean I reviewed the code carefully, though. :P So let's just make sure there are reviews before merging, specially because it's about decryption. :)

dominictarr:
this is probably a small performance regression, and on that note, @arj03 is probably the person best able to review this.

also note: Also, with this in place, i might be able to prototype private groups quite easily.
But currently, the decryption occurs in the _decode_ function. This isn't really right, because decode should be a pure function... this was just the easiest place to add it. I think I'd rather refactor this a bit though.

hmm, another idea would be to add assertions to check that private messages are never returned decrypted unless private was true, then use it for a while (but don't make it the default npm version)

minimal.js
```diff
@@ -46,7 +79,7 @@ module.exports = function (dirname, opts) {
     state.queue = []
     append(batch, function (err, v) {
       batch.forEach(function (data) {
-        db.post.set(data)
+        db.post.set(u.rebox(data))
```
arj03: u is not defined here

indexes/clock.js
```diff
@@ -56,7 +55,7 @@ module.exports = function (db, opts) {
       opts.keys = false
       opts.values = true
 
-      return pull(index.read(opts), Format(keys, values))
+      return pull(index.read(opts), Format(keys, values, opts.private === true))
```
arj03: Format is not defined

arj03:
The test found those two. Im also getting a

/home/chrx/dev/secure-scuttlebutt/node_modules/ssb-validate/index.js:215
  if(err) throw err
          ^

Error: encoded message must not be larger than 8192 bytes
    at exports.isInvalidShape (/home/chrx/dev/secure-scuttlebutt/node_modules/ssb-validate/index.js:70:12)
    at Object.exports.create (/home/chrx/dev/secure-scuttlebutt/node_modules/ssb-validate/index.js:214:13)
    at /home/chrx/dev/secure-scuttlebutt/minimal.js:135:17

arj03:
I ran the benchmarks on this (published in the bench-ssb-share repo) and the flume test (3) is much slower. Going from 12s to 17s. The rest seems to be within standard deviation, maybe a little bit slower.

dominictarr:
oops, sorry @arj03 I had those fixes on my repo but hadn't checked them in yet. fixed now.

Hmm, if the flume benchmark is slower, it's not because of this code (because that benchmark doesn't even `require('secure-scuttlebutt')`) it might be because of changes to ssb-validate though... hmm, having trouble reproducing this perf difference

dominictarr:
hmm, actually 03-flume only takes me 6 seconds! I'm surprised since I have an old slow computer... 
on node 6 it took 7 seconds, node 8 it took 6

arj03:
You're right. My deps got updated and it appears that something in pull-stream 3.6.2 is a lot slower. I've tried 3.6.2, 3.6.7 (both slow) but 3.6.1 is faster.

... and yeah my computer is slow :smiley:

arj03:
Its really strange. I ran the pull-stream benchmarks on 3.6.1 and 3.6.7 and run time is the same, but the heap size is much better in the node test on 3.6.7 so I don't understand why it should affect the flume benchmark like is does.

evbogue:
This is cool, as we wouldn't be decrypting messages over and over again, wasting lots of cpu. However, I am cautious about change because of the possibility of leaking decrypted data to the public, as @dominictarr mentioned:

>  It would be okay if local clients such as patchwork got decrypted messages, but it could be a leak for web viewers.

As you mentioned, this makes a lot of sense for Patchwork. But let's say I'm using Decent with a remote key that is different than the server. Where does my private index get stored if my key is different than the pub I'm using?

Right now Minbase uses the Patchbay@6 code to decode private messages within the browser. I think this means the server has zero knowledge of what is in those messages.

dominictarr:
@evbouge if the sever has a different key, this isn't gonna effect you at all. You'll still need to build your own indexes (or scan the private messages and attempt decryption) on the client side, and this doesn't stop you from doing that.

extras.js
```diff
@@ -12,7 +11,7 @@ module.exports = function (db, config, keys) {
       return [data.timestamp]
     }))
     .use('feed', require('./indexes/feed')())
-    .use('links', require('./indexes/links')(keys))
+    .use('links', require('./indexes/links')())
```
arj03: This change doesn't look correct does it? keys is still used in links.

arj03:
I'm testing this with patchbay but it crashes in some internal code because I get messages that are the value false. Havn't exactly figured out the cause yet. And then there are the [test failing](https://github.com/ssbc/secure-scuttlebutt/pull/199#issuecomment-379880898) + the links change that looks a bit strange.

I'll be testing this with ssb-viewer as well but have to run to a meeting now. Just wanted to leave my comments so I don't forget.

evbogue:
Ok, I get it. :+1: for a contiued dedication to websbot. 

On my end, these decentralized public nodes are just a stopgap towards a fully distributed peer in the browser.

Just thinking out loud: maybe I could generate a flumedb log in the browser and then save it to localStorage?

arj03:
The test crash is because ssb-validate 3.0.7 now throws on invalid input. The following patch fixes that:

diff --git a/minimal.js b/minimal.js
index e657899..fd675a4 100644
--- a/minimal.js
+++ b/minimal.js
@@ -132,12 +132,19 @@ module.exports = function (dirname, keys, opts) {
     })
   })
   db.append = wait(function (opts, cb) {
-    var msg = V.create(
-      state.feeds[opts.keys.id],
-      opts.keys, opts.hmacKey || hmac_key,
-      opts.content,
-      timestamp()
-    )
+    try {
+      var msg = V.create(
+        state.feeds[opts.keys.id],
+        opts.keys, opts.hmacKey || hmac_key,
+        opts.content,
+        timestamp()
+      )
+    }
+    catch (err) {
+      cb(err)
+      return
+    }
+
     queue(msg, function (err) {
       if(err) return cb(err)
       var data = state.queue[state.queue.length-1]

arj03:
I've tested ssb viewer and sbot.createUserStream({ id: feedId }) returns nothing. But sbot.query.read works fine.

minimal.js
```diff
@@ -28,9 +44,26 @@ function toKeyValueTimestamp(msg) {
...
-module.exports = function (dirname, opts) {
+function isString (s) {
+  return 'string' === typeof s
+}
+
+module.exports = function (dirname, keys, opts) {
   var hmac_key = opts && opts.caps && opts.caps.sign
+
+  var codec = {
+    encode: function (obj) {
```
arj03: I'm getting some:

could not retrive msg: NotFoundError: Key not found in database [@5yP0SoZVkMIXUQcL7nPGsw6yYIxE23UBd+rLO+6OSWc=.ed25519,5]

Shouldn't this be just JSON.stringify? I can understand overwriting decode.

arj03: This seems to fix the createUserStream bug.

dominictarr:
@arj03 

> Ouch. This allows one to read private messages of the machine running ssb-viewer.

this is exactly what I'm trying to avoid. current ssb-viewer should never use `private: true` as an option, and ssb should never return decrypted data unless private: true is passed.

dominictarr:
@arj03 okay, I found it. fixed in d3989790c984d9581147c77e44547950dfc67d65

arj03:
Excellent! I'll do some more testing, what about the [codec change](https://github.com/ssbc/secure-scuttlebutt/pull/199#discussion_r181232640)?

dominictarr:
also, made PR to fix some privacy related problems in ssb-viewer `%3GamV50bdvjzA+iDbCPZ24Iu6NMP1OCwfJE4RXTOExU=.sha256`

dominictarr:
it probably really should have the newlines, because that is what ssb-keys uses
https://github.com/ssbc/ssb-keys/blob/master/index.js#L127
if we move towards that, we can have a design were we don't actually rely on JSON string ordering... by always keeping the original string and never serializing again... it might bloat the output a little.. but that is less embarrasing mistake than relying on the order

dominictarr:
this was feeling just about ready to merge... then I asked myself: "is there anything else this could affect"
and I realized: git-ssb, git-ssb has some support for secret repos. @clehner what do you think?

PR #203, opened 2018-05-02
Fix signature hash not working when reboxing

arj03:
With this ssb-validate doesn't fail on the first message. It fails after around 1723 so there is still a bug in here somewhere.

### DIFF
--- a/util.js
+++ b/util.js
@@ -41,15 +41,17 @@ exports.wait = function () {
 }
 
 var reboxValue = exports.reboxValue = function (value, isPrivate) {
-  return isPrivate === true ? value : {
-    previous: value.previous,
-    author: value.author,
-    sequence: value.sequence,
-    timestamp: value.timestamp,
-    hash: value.hash,
-    content: value.cyphertext || value.content,
-    signature: value.signature
+  if (isPrivate === true) return value
+
+  var o = {}
+  for (var key in value) {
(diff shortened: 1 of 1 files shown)


arj03:
Ah it looks like git-ssb changes the order around so that author is first.

arj03:
Now it validates my complete log. This should fix https://github.com/ssbc/scuttlebot/issues/495

dominictarr:
oh dang, I didn't even realize that git-ssb was doing that. that's gonna make upgrading the protocol a little more complicated, so I'm glad to know that now! I'm gonna merge this right away because it obviously is urgent... this needs test coverage too. Can you link me to were git-ssb does that?

dominictarr:
actually, I gotta run off right now, won't be able to merge for a few hours, but I added you as npm owner @arj03 so please go a head merge and publish.

cryptix:
Oh, synchronicity! :)

I also stumbled over this field ordering change but couldn't figure out how it was introduced. 

On my feed it occurred 6 month ago, though.. See `%fOyyo01vfF5DESP/sMWiNZNzM2U/DwmUneY0EQ0F0NY=.sha256`

arj03:
This now has tests aba89d3754b3c09ca262066e0fc4600c30c8b75b

PR #204, opened 2018-05-03
Bring level up-to-date

christianbundy:
Sorry for the never-ending PRs.

This (like the others) is meant to circumvent https://github.com/Level/leveldown/issues/455 by upgrading `level` to the most recent version. It passes `npm test`, but skepticism is warranted.

Thanks for putting up with me! :wink:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2325 @@
+{
+  "name": "secure-scuttlebutt",
+  "version": "18.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abstract-leveldown": {
+      "version": "2.6.3",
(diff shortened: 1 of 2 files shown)


myf:
level-test also needs to be updated eventhough it's a dev dependency. building secure-scuttlebutt itself will break it

vweevers:
Do your tests cover usage of `level-sublevel`? Curious to know if that's compatible with latest level.

christianbundy:
@vweevers Negative. I'm not very familiar with any of the Level components, so any/all expertise is very appreciated.

christianbundy:
Looks like this waas taken care of in 78de3f624dd7718d606b66e95c1de35edd01327e, thanks!

PR #206, opened 2018-05-10
Upgrade to level-test@^3.0.0

christianbundy:
This should be one of the last changes for Node 10 compatibility across the scuttleverse.

WIP: Currently fails NPM, will update as that changes.


dominictarr:
oh, this was fixed in 78de3f62 but thanks! 18.0.2 should be full node 10 compat!

PR #211, opened 2018-05-24
Don't use while + shift pattern 

arj03:
as that can be quite slow if the array is large


staltz:
Good idea. However, what's the purpose of the slice? If `n = arr.length`, then `arr.slice(n)` would be just `[]`

dominictarr:
yup, merging. I am happy that the concurrency logic is correct.

dominictarr:
merged into 18.0.6

PR #213, opened 2018-06-15
Support unbox with multiple ssb-identities

theproductiveprogrammer:


evbogue:
What is the use-case for this pull-request. What are you trying to accomplish with it, and what project is this needed for?

It seems to me you'd be able to try passing different keypairs to unbox client-side.

dominictarr:
context: I have been advising @theproductiveprogrammer on how to add support for multilpe identities. @evbogue is right, it's not just me who looks at PRs, please include a description of what you are trying to do, as well as how it works. as an example: here is an example PR that I made: https://github.com/ssbc/secure-scuttlebutt/pull/199 (though this PR doesn't require that much explaination I think)

dominictarr:
is `addl` short for additional? I think it's okay to use abbreviations in function arguments (easy to change) but in apis is a different matter.

theproductiveprogrammer:
@evbogue I am trying to add support for multiple identities managed by a single scuttlebot instance. This means that a single sbot server will be create and manage feeds for multiple clients (with the [ssb-identities](https://github.com/ssbc/ssb-identities) plugin)

The project for which I want this support is [Everlife.Ai](http://everlife.ai/). In this case, we need to support 'hosting' of your scuttlebot node (similar to hosting your blog) for many of our users who do not want to set up their own nodes. In this case the sever sbot acts as a "bot herder" shepherding the identities. Of course they are free to move out to another host or to self-hosting at any time.

@dominictarr would probably know more about other projects & usecases where multiple identities are useful

theproductiveprogrammer:
Yes `addl` was short for `additional`. As an API would `setAdditionalKeysArray` work? Another option would be to add/remove each additional key (rather than an array reference) - `addAdditionalKey/rmAdditionalKey` if we think making it more explicit would be better.

evbogue:
Nice, I get why you need this for your project now. Thanks for taking the time to fill me in @theproductiveprogrammer.

staltz:
@theproductiveprogrammer how about `addAdditionalKeys`?

theproductiveprogrammer:
Ok the best I've come up with is `addIdentity/rmIdentity` and updated the PR.

theproductiveprogrammer:
I think that's an excellent idea! I'll update the PR's to `addUnboxer`'s

theproductiveprogrammer:
@dominictarr I've pushed the `addUnboxer` API and also added a commit for replication.

dominictarr:
Okay, this is good. I'm gonna merge this, but I'm gonna move the unboxers array off the codec object. The codec is ment to be one of these thing: https://github.com/level/codec#encoding-format So it should just be that.

dominictarr:
merged into 18.1.0

PR #214, opened 2018-06-24
BUGFIX: Crashing on `addUnboxer`

theproductiveprogrammer:
[This commit](https://github.com/ssbc/secure-scuttlebutt/commit/281ae35e755b44452bd5db395cd10ba0867756f4) moved unboxer out of the `codec`. This caused a crash in `addUnboxer` so fixed it.


dominictarr:
oh, can you make a test, so this doesn't happen again?

PR #215, opened 2018-08-05
feed index: use resolved timestamp to ensure unwanted items can't get pinned to top of feed

mmckegg:
### ⚠️ Because scuttlebutt is a _"they who said somebody should, should"_ project, if nobody complains about this before _19 August_, I will be merging it! 

So please speak up if this is going to break your code 🐵 

@dominictarr @mixmix @evbogue @staltz @soapdog @arj03 @clehner 

---

There is currently a massive problem with `getFeedStream` that basically renders it unusable. It's that if someone posts a message with a timestamp far in the future, it will get permanently stuck at the top of your feed. This limitation means that I haven't been using this index at all, and instead am forced to make my own index.

**This PR solves the problem by indexing `Math.min(msg.timestamp, msg.value.timestamp)` to ensure that a message can never be newer than the time it was synced.** It also adds a new `rts` key to the results (alongside `timestamp`) so that you can easily get this value. This brings `getFeedStream` inline with ssb-backlinks.

Technically this is a breaking change, as clients may be using `msg.value.timestamp` to resume the feed with a stepper (endless scrolling). However, in practice, I don't think that it will break any clients as messages with `value.timestamp > timestamp` are very rare, and the client can easily be fixed by using `msg.rts || msg.value.timestamp`. 

If no one else thinks that this is worth merging (because of the breaking change), I will continue to use my custom index in patchwork (and we'll be left with one more useless legacy index [that no one should use] in scuttlebutt).

Another good talking point might be whether the key should be `rts` or something else. I chose this as that is what `ssb-backlinks` uses.

### DIFF
--- a/indexes/feed.js
+++ b/indexes/feed.js
@@ -1,6 +1,5 @@
 'use strict'
 var pull = require('pull-stream')
-var path = require('path')
 var ltgt = require('ltgt')
 var u = require('../util')
 
@@ -8,8 +7,8 @@ var ViewLevel = require('flumeview-level')
 
 module.exports = function (db) {
 
-  var createIndex = ViewLevel(2, function (data) {
-    return [[data.value.timestamp, data.value.author]]
+  var createIndex = ViewLevel(3, function (data) {
+    return [[Math.min(data.timestamp, data.value.timestamp), data.value.author]]
   })
 
   return function (log, name) {
@@ -24,25 +23,21 @@ module.exports = function (db) {
 
       var keys = opts.keys
       var values = opts.values
-      opts.keys = false
+      opts.keys = true
       opts.values = true
 
-      return pull(index.read(opts), u.Format(keys, values, opts.private === true))
+      return pull(
+        index.read(opts),
+        pull.through(item => {
+          if (item.value && item.key) {
+            // make resolved timestamp available
+            item.value.rts = item.key[0]
+          }
+        }),
+        u.Format(keys, values, opts.private === true)
+      )
     }
 
     return index
-
   }
 }
-
-
-
-
-
-
-
-
-
-
-
-
--- a/util.js
+++ b/util.js
@@ -60,7 +60,8 @@ var reboxValue = exports.reboxValue = function (value, isPrivate) {
 var rebox = exports.rebox = function (data, isPrivate) {
   return isPrivate === true ? data : {
     key: data.key, value: reboxValue(data.value, isPrivate),
-    timestamp: data.timestamp
+    timestamp: data.timestamp,
+    rts: data.rts
   }
 }
 


evbogue:
I approve of this commit being merged.

Serkan-devel:
> "they who said somebody should, should" project
What do you mean with that?

staltz:
I just experienced this issue this weekend because someone commented on a thread and it's timestamped 5 months from now. I saw it in MMMMM, which is using ssb-threads which as far as I remember is using feedStream

Serkan-devel:
@staltz I've tried MMMMM on an S6 and it doesn't load any page. The hamburger menu is also unresponsive.
Have you thought of implementing [Sentry](https://sentry.io/) to your app?
It might be easy for error reporting

cryptix:
All for fixing things but please help me for a second to understand the issue better. I'm not sure I've seen this before. If a user writes _a message from the future_ (no rtc, selected wrong timezone, ...), the output of `getFeedStream` for that person would be broken until they would start sending messages from after that point?

staltz:
@Serkan-devel You could post issues on the mmmmm repo and also your experience depends which version of the app you were running.

mmckegg:
@Serkan-devel 

>> "they who said somebody should, should" project
>
> What do you mean with that?

I am suggesting that the person who identifies a problem should be the person that fixes it. "Somebody should do the dishes".

But really what I am saying is that I am sick of waiting on other people to approve pull requests in ssbc as it it becomes a huge bottleneck in development, and instead am going to try and take ownership of getting this issue merged! Hence my engagement baiting threat at the top 😆 

---

@staltz 

> Is the breaking change experienced simply by the order being a bit different?

Yes. Say you were using a stepper function to iterate over the feed in reverse, it is possible that it might end on a message that has a `value.content.timestamp` that is more recent than the post before it (or even the starting post). This would cause the stepper to get stuck in a loop, repeating the same posts over and over. But this would only happen if there was a post that would currently render in the wrong place anyway, and is easy for the client maintainer to solve by updating their stepper function to use `msg.rts` instead.

---

@cryptix 

> ... please help me for a second to understand the issue better ... If a user writes a message from the future (no rtc, selected wrong timezone, ...), the output of getFeedStream for that person would be broken until they would start sending messages from after that point?

This fix is mostly for people with their clocks set correctly so that people with incorrectly set clocks don't spoil their feed. It would only be a problem for someone with an incorrectly set clock on initial sync and then later set their clock correctly (as it might cause posts dated after time at sync to be rendered in the wrong order).

What happens right now if someone posts a message with a future timestamp, is that it will get inserted at that future point in the index, and all other posts will appear before it. If the feed is being retrieved in reverse (newest first), it will cause that post to get stuck at the top for everyone else.

---

Thanks everyone for the feedback 😄

mixmix:
Thanks for the heads up and the breakdown of the situation matt.
Sounds totally reasonable

dominictarr:
@mmckegg go for it: but if you want to cut a new release of secure-scuttlebutt, please link it into scuttlebot and run scuttlebot's tests too. If that passes, go for it.

mmckegg:
Okay, cool! I will do some ssb stuff next week ⛵️

mmckegg:
Linked to scuttlebot and all tests passed! Gonna merge this (finally!)

PR #218, opened 2018-09-18
Prevent crash when unboxer throws exception

### LINKED ISSUE #217: Invalid `content` property throwing errors
I think I've caused/found a failure mode by setting the property `"content": "&sbBmsB(...).sha256"` as mentioned [in this thread](https://github.com/flumedb/flumelog-offset/issues/13#issuecomment-422475518).  I'm currently unclear on the severity of this bug, or whether my experiment is causing problems for others on the network. I'm getting two errors on master:

```
Error: public_key must be a buffer of size crypto_scalarmult_BYTES
    at exports.crypto_scalarmult (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/sodium-chloride/index.js:118:8)
    at Object.exports.decrypt.exports.multibox_open (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/private-box/index.js:60:16)
    at exports.unbox (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/ssb-keys/index.js:157:16)
    at Array.main_unboxer (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:60:49)
    at unbox (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:21:36)
    at Object.decode (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:68:14)
    at /home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/flumelog-offset/inject.js:45:22
    at /home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/flumelog-offset/frame/recoverable.js:48:11
    at /home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:74:11
    at get (/home/christianbundy/.node_modules/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:29:7)
```

```
Error: bad p size
    at Object.nacl.scalarMult (/home/christianbundy/Source/patchwork/node_modules/tweetnacl/nacl-fast.js:2202:51)
    at exports.crypto_scalarmult (/home/christianbundy/Source/patchwork/node_modules/sodium-browserify-tweetnacl/index.js:76:31)
    at Object.exports.decrypt.exports.multibox_open (/home/christianbundy/Source/patchwork/node_modules/private-box/index.js:60:16)
    at exports.unbox (/home/christianbundy/Source/patchwork/node_modules/ssb-keys/index.js:157:16)
    at Array.main_unboxer (/home/christianbundy/Source/patchwork/node_modules/secure-scuttlebutt/minimal.js:60:49)
    at unbox (/home/christianbundy/Source/patchwork/node_modules/secure-scuttlebutt/minimal.js:21:36)
    at Object.decode (/home/christianbundy/Source/patchwork/node_modules/secure-scuttlebutt/minimal.js:68:14)
    at /home/christianbundy/Source/patchwork/node_modules/flumelog-offset/inject.js:76:29
    at /home/christianbundy/Source/patchwork/node_modules/flumelog-offset/frame/recoverable.js:48:11
    at /home/christianbundy/Source/patchwork/node_modules/aligned-block-file/blocks.js:74:11
```

I think we may be able to resolve this bug by wrapping [`ssbKeys.unbox()`](https://github.com/ssbc/ssb-keys/blob/2d9c2dd96b5c94e3d73bad41884fc6a31be7e4bc/index.js#L153-L162) in a `try...catch` block, or wrapping it [here](https://github.com/ssbc/secure-scuttlebutt/blob/beff8d063a20a1604ffe9110eb0d0f0911fc2e95/minimal.js#L21) so that unbox errors can be handled without anything catching fire.

christianbundy:
This resolves #217 by catching unboxer exceptions and moving on as if nothing happened. This does *not* log uncaught exceptions, and instead leaves that to the unboxer. Please let me know if this needs changes before merging or if we should instead handle this elsewhere. Thanks!

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -18,7 +18,11 @@ var isArray = Array.isArray
 function unbox(data, unboxers) {
   if(data && isString(data.value.content)) {
     for(var i = 0;i < unboxers.length;i++) {
-        var plaintext = unboxers[i](data.value.content, data.value)
+        var plaintext
+        try {
+          plaintext = unboxers[i](data.value.content, data.value)
+        } catch (_) { }
+
         if(plaintext) {
             data.value.cyphertext = data.value.content
             data.value.content = plaintext


dominictarr:
hmm, well the unboxers shouldn't really throw, so this would also mean that some problems with unboxers would go undetected, until people notice things being undecrypted that shouldn't. I think we probably should log the error here, because it is an unexpected error, but not a fatal one.

Hmm, after fixing the error, you'd need to rebuild all the indexes... there isn't a non-manual way to do that from the unboxers point of view.

dominictarr:
... although, for private-groups we'll need to sometimes rebuild all the indexes, really just reindex the messages which can now be decrypted, and there is probably a clever way to do that, but in the short term it will be easier to just reindex everything.

PR #219, opened 2018-09-20
Async unboxer

christianbundy:
This PR depends on [asynchronous codecs](https://github.com/flumedb/flumelog-offset/pull/14) and makes a few changes to allow asynchronous unboxers:

- Unboxers get passed `value` rather than `content, value`
  - This allows unboxers to set props like `private: true`
  - Unboxers can use `value.content` rather than a dedicated `content` argument
- Unboxers are currently required to be `async function`s that return promises
  - I don't have any strong opinions on this, it was just the most consistent implementation.
  - These promises are processed in a chain, where one unboxer's result is passed to the next
  - This allows for combinations like `main_unboxer(blob_unboxer(value))`
- Unboxers return `false` to indicate that they don't want to process a value
  - This *shouldn't* conflict with any other values, unless an unboxer wants to return `content: false`
  - The performance gains are probably trivial, but this means we're only returning `value` when necessary

I'm sure this will be the first of a few iterations, please let me know if you have any suggestions or see any pain points that we should try to avoid. Thanks for taking a look!

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -18,13 +18,10 @@ var isArray = Array.isArray
 function unbox(data, unboxers) {
   if(data && isString(data.value.content)) {
     for(var i = 0;i < unboxers.length;i++) {
-        var plaintext = unboxers[i](data.value.content, data.value)
-        if(plaintext) {
-            data.value.cyphertext = data.value.content
-            data.value.content = plaintext
-            data.value.private = true
-            return data
-        }
+      var plaintext = unboxers[i](data.value)
+      if(plaintext) {
+        data.value = plaintext
+      }
     }
   }
   return data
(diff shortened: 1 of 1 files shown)


dominictarr:
unboxers should just return a value, not set be mutating the message. also, the bit that calls the unboxers already set `private: true`

what is the rationale for not returning after the content?

hmm, you know, really I think putting this into the unboxers is the wrong place. it's not really decrypting anything, it's looking something up. what would be more elegant: is an async map that happens after reading a value, that is not part of the codec.

Hmm, maybe a good place to put this is in `flumedb`... it already wraps the streams, it could just wrap the log's stream method and get method.

christianbundy:
> also, the bit that calls the unboxers already set private: true

I removed this because `private: true` didn't seem relevant to a blob "unboxer", and I thought that different unboxers might want to make their own changes to `data.value` rather than just `data.value.content`.

> what is the rationale for not returning after the content?

It's asynchronous, so I wanted to avoid having to check `content === data.value.content` to decide which results could be ignored and which needed to be merged into the Franken-object.

> Hmm, maybe a good place to put this is in flumedb... it already wraps the streams, it could just wrap the log's stream method and get method.

Cool, I'll start reading about flumedb and see whether it's something I'm able to do. I totally agree that's *way* more elegant. Time to deep-dive into flumedb. Thanks for the feedback!

dominictarr:
btw, for encrypted blob content, I'd recommend encrypting the message like normal, and then linking to the blob with an unbox key, just a generalization of how it's done with blobs currently.

PR #222, opened 2018-09-27
Add `addMap()` method

christianbundy:
I'm working toward using the flumedb map as a Scuttlebot plugin, but it seems like the first step would be to accept a function as `opts.map` that gets passed to flume. Is this the best way to go about that change?

Backlink: https://github.com/flumedb/flumedb/pull/19.

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -108,7 +108,7 @@ module.exports = function (dirname, keys, opts) {
 
   //NOTE: must use db.ready.set(true) at when migration is complete
   //false says the database is not ready yet!
-  var db = Flume(log, false, maps ? chainMaps : null)
+  var db = Flume(log, false, maps.length ? chainMaps : null)
   .use('last', require('./indexes/last')())
 
   var state = V.initial(), ready = false
--- a/minimal.js
+++ b/minimal.js
@@ -93,9 +93,22 @@ module.exports = function (dirname, keys, opts) {
 
   var log = OffsetLog(path.join(dirname, 'log.offset'), {blockSize:1024*16, codec:codec})
 
-  //NOTE: must use db.ready.set(true) at when migration is complete
+  const maps = []
+  const chainMaps = (val, cb) => {
+    let idx = -1 // haven't entered the chain yet
+    const next = (err, val) => {
+      idx += 1
+      if (err || idx === maps.length)
+        cb(err, val)
+      else
+        maps[idx](val, next)
+    }
+    next(null, val)
+  }
 
-  var db = Flume(log, false, opts.map) //false says the database is not ready yet!
+  //NOTE: must use db.ready.set(true) at when migration is complete
+  //false says the database is not ready yet!
+  var db = Flume(log, false, maps ? chainMaps : null)
   .use('last', require('./indexes/last')())
 
   var state = V.initial(), ready = false
@@ -210,6 +223,9 @@ module.exports = function (dirname, keys, opts) {
 
   db.unbox = function (data, key) {
     return unbox(data, unboxers, key)
+  }
+  db.addMap = function(fn) {
+    maps.push(fn);
   }
 
   return db
--- a/minimal.js
+++ b/minimal.js
@@ -95,17 +95,22 @@ module.exports = function (dirname, keys, opts) {
 
   const maps = []
   const chainMaps = (val, cb) => {
-    if (!maps.length) return cb(null, val)
-
-    let idx = -1 // haven't entered the chain yet
-    const next = (err, val) => {
-      idx += 1
-      if (err || idx === maps.length)
-        cb(err, val)
-      else
-        maps[idx](val, next)
+    const mapCount = maps.length
+    if (!mapCount) {
+      return cb(null, val)
+    } else if (mapCount === 1) {
+      maps[0](val, cb)
+    } else {
+      let idx = -1 // haven't entered the chain yet
+      const next = (err, val) => {
+        idx += 1
+        if (err || idx === maps.length)
+          cb(err, val)
+        else
+          maps[idx](val, next)
+      }
+      next(null, val)
     }
-    next(null, val)
   }
 
   //NOTE: must use db.ready.set(true) at when migration is complete
--- a/minimal.js
+++ b/minimal.js
@@ -95,7 +95,7 @@ module.exports = function (dirname, keys, opts) {
 
   //NOTE: must use db.ready.set(true) at when migration is complete
 
-  var db = Flume(log, false) //false says the database is not ready yet!
+  var db = Flume(log, false, opts.map) //false says the database is not ready yet!
   .use('last', require('./indexes/last')())
 
   var state = V.initial(), ready = false
--- a/minimal.js
+++ b/minimal.js
@@ -95,6 +95,8 @@ module.exports = function (dirname, keys, opts) {
 
   const maps = []
   const chainMaps = (val, cb) => {
+    if (!maps.length) return cb(null, val)
+
     let idx = -1 // haven't entered the chain yet
     const next = (err, val) => {
       idx += 1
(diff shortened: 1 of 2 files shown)


minimal.js
```diff
@@ -93,9 +93,22 @@ module.exports = function (dirname, keys, opts) {
...
+      else
+        maps[idx](val, next)
+    }
+    next(null, val)
+  }
 
-  var db = Flume(log, false) //false says the database is not ready yet!
+  //NOTE: must use db.ready.set(true) at when migration is complete
+  //false says the database is not ready yet!
+  var db = Flume(log, false, maps ? chainMaps : null)
```
dominictarr: this is always gonna eval to `chainMaps` because `maps` is a const. I guess this is left over from when maps was a constructor arg.

christianbundy: Thanks! I think it's an issue where `!![] === true`, I'll push a fix now to verify that we're checking `maps.length` instead.

christianbundy: The behavior should look like this now:

```js
const maps = []

// no functions added to `maps` array
!!maps.length // => false

// function added to `maps` array
maps.push(x => 42)
!!maps.length // => true
```

minimal.js
```diff
@@ -93,9 +93,22 @@ module.exports = function (dirname, keys, opts) {
...
+    let idx = -1 // haven't entered the chain yet
+    const next = (err, val) => {
+      idx += 1
+      if (err || idx === maps.length)
+        cb(err, val)
+      else
+        maps[idx](val, next)
+    }
+    next(null, val)
+  }
```
christianbundy: There's also a cleaner implementation if you're open to having [`run-waterfall`](https://github.com/feross/run-waterfall) added to the project:

```js
function chainMaps(value, done) {
  waterfall([
    cb => cb(null, value),
    ...maps,
    done
  ]
}
```

Is this worth adding another package, or would you prefer to keep the current implementation?

Suggested here: `%RSVrlUQcZSmkmXOqMSfr0XTKeCHnFPHyWkKmU5LKSaY=.sha256`

dominictarr:
I ran the bench marks, and didn't detect a difference. can you add documentation and then merge and publish?

minimal.js
dominictarr: i mean shouldn't it just be `chainMaps` with no `?` at all? because Flume is initialized before the ssb constructor returns, so maps will _always_ be empty at that point... because no one has had a chance to call `addMaps`

Is there another module that's just like run-waterfall but only passes a single argument, not a variable number?

christianbundy: ...yep, looks like this wasn't even calling `chainMaps` since most of my testing was with a hardcoded `maps` array. I've just pushed another commit to fix this, but it looks like it adds a 5-10% slowdown across the board. I'm going to work on that now.

I don't think so, but it looks like this is the same algorithm that run-waterfall is using (sans the multiple arguments). I can rename the variables to match what run-waterfall is using if you think that might be useful from a documentation/readability perspective.

minimal.js
```diff
@@ -93,9 +93,24 @@ module.exports = function (dirname, keys, opts) {
 
   var log = OffsetLog(path.join(dirname, 'log.offset'), {blockSize:1024*16, codec:codec})
 
-  //NOTE: must use db.ready.set(true) at when migration is complete
+  const maps = []
+  const chainMaps = (val, cb) => {
+    if (!maps.length) return cb(null, val)
```
christianbundy: I was thinking I'd be able to squeeze out a few milliseconds of performance by checking a boolean here instead, but [my benchmarks](https://jsperf.com/arrboolcheck) are showing that `if (!maps.length)` is just as fast as checking `if (hasMaps)` or similar.

minimal.js
christianbundy: Here are the results of ten tests on with the original code (mean 18.9952):

```
random-read 1000000 0 51631.55720776539 19.368
random-read 1000000 0 48334.863937358015 20.689
random-read 1000000 0 52540.324699206634 19.033
random-read 1000000 0 51435.03754757741 19.442
random-read 1000000 0 54174.11560756271 18.459
random-read 1000000 0 53341.86803221849 18.747
random-read 1000000 0 51808.10278727593 19.302
random-read 1000000 0 54051.132371223175 18.501
random-read 1000000 0 54232.87596941266 18.439
random-read 1000000 0 55642.10994880925 17.972
```

And the results of the new code (mean 19.2827):

``` 
random-read 1000000 0 49798.31681689159 20.081
random-read 1000000 0 53250.9718302359 18.779
random-read 1000000 0 53584.82477762298 18.662
random-read 1000000 0 51371.622315832734 19.466
random-read 1000000 0 52938.06246691371 18.89
random-read 1000000 0 50900.94675760969 19.646
random-read 1000000 0 52260.256075254765 19.135
random-read 1000000 0 51810.787005854625 19.301
random-read 1000000 0 51250.51250512505 19.512
random-read 1000000 0 51666.236114699044 19.355
```

It looks like this takes only 1.5% longer, so the performance change seems much smaller than I originally thought.

dominictarr: interesting, what if you had a special case that skipped the array if there was only 1 map?

christianbundy: Benchmarks on the new code seem to show that 0-1 maps seem to be within the standard error of the mean (maybe?) or I'm totally screwing something up. I decided to cast to a variable rather than check `maps.length` twice, but if checking `!maps.length` and `maps.length === 1` is the right call I'm happy to do that instead.

No maps (mean 19.3461):

```
random-read 1000000 0 52946.471117700006 18.887
random-read 1000000 0 50965.80194689364 19.621
random-read 1000000 0 52675.93763168984 18.984
random-read 1000000 0 50640.60363599534 19.747
random-read 1000000 0 53533.19057815846 18.68
random-read 1000000 0 52898.857384680494 18.904
random-read 1000000 0 52681.48772521336 18.982
random-read 1000000 0 51559.68032998196 19.395
random-read 1000000 0 52617.73217574323 19.005
random-read 1000000 0 48271.86715582159 20.716
```

1 map (mean 19.0712):

```
random-read 1000000 0 52222.04814872839 19.149
random-read 1000000 0 51368.98340781836 19.467
random-read 1000000 0 52887.66659614977 18.908
random-read 1000000 0 51797.36869367036 19.306
random-read 1000000 0 52770.44854881267 18.95
random-read 1000000 0 54019.014693171994 18.512
random-read 1000000 0 52653.74894692502 18.992
random-read 1000000 0 52868.09410520751 18.915
random-read 1000000 0 52910.05291005292 18.9
random-read 1000000 0 50986.59052669148 19.613
```

2 maps (mean 19.6632):

```
random-read 1000000 0 52143.08061320263 19.178
random-read 1000000 0 51109.066748441175 19.566
random-read 1000000 0 50092.67144216801 19.963
random-read 1000000 0 48721.071863581004 20.525
random-read 1000000 0 49026.81766926509 20.397
random-read 1000000 0 52026.42942614848 19.221
random-read 1000000 0 51458.85864251531 19.433
random-read 1000000 0 51671.57546633596 19.353
random-read 1000000 0 54106.69840926307 18.482
random-read 1000000 0 48747.19703617042 20.514

```

christianbundy:
Based on [these benchmarks](https://github.com/ssbc/secure-scuttlebutt/pull/222#discussion_r222475310) I think the performance is reasonable again. I've added some documentation, please let me know if you'd like any other changes made. Thanks!

dominictarr:
great, it i'm not so picky about internal style things, unless there is a performance reason or something. (external apis are worth arguing about though!) merging!!!

dominictarr:
I doubt this actually broke anything, but it's technically a breaking change so published as 13.0.0
also, update to secure-scuttlebutt@18.5 so get addMap by @christianbundy (a key step towards private-groups, although his reason for it was to experiment with blobs-as-content)

dominictarr:
merged into 18.5.0, and scuttlebot@13 !

PR #223, opened 2018-10-09
Reuse hash of message instead of recomputing it

arj03:
This reuses the id of ssb-validates append instead of recomputing it. On my machine this saves 5 seconds out of 157 seconds for benchmark 11 in bench-ssb.

### DIFF
--- a/db.js
+++ b/db.js
@@ -1,5 +1,4 @@
 var path = require('path')
-var ViewLevel = require('flumeview-level')
 var ViewHashTable = require('flumeview-hashtable')
 
 module.exports = function (dir, keys, opts) {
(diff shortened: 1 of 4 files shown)


arj03:
There was one more, so the improvement is now a bit bigger. I would like some eyes on the second one, as I'm not 100% its good as the function seems to buffer. The tests parses.

dominictarr:
I think it's not safe to take the ID from `state.feeds[feed.id].id` it's _possible_ that multiple messages are appended to `ssb-validate`'s queue before feed is called - that is a supported and tested feature of ssb-validate. A better fix might be to change ssb-validate to use `{key, value, timestamp}` format in it's output queue.

dominictarr:
I can just see some bug that is introduced because something, possibly some change, happens that ssb-validate queues several messages before they get passed to ssb-validate. This would mean messages get a completely wrong key!

arj03:
Right, I removed the last commit as that was not good.

PR #228, opened 2018-10-17
Move unboxer metadata from `msg.value` to `msg.value.meta`

### LINKED ISSUE #225: using addMap() seems to break replication
I'm unsure whether this is a problem with secure-scuttlebutt or scuttlebot, but using the `addMap()` function seems to stop my messages from being synced to other peers. My intuition is that Scuttlebot might be hashing (or broadcasting?) the mapped data rather than the original, but I'm not familiar enough to understand exactly what's going on.

---

If the above *is* the case, maybe we'd need a way for critical message infrastructure like replication/verification to opt out of any message modifications?

christianbundy:
This builds on #227 and resolves #225 in a backward-compatible way.

## TL;DR

This standardizes on `msg.value.meta` being the place for unboxers and map functions to create metadata that **will not break replication**. It also standardizes on `msg.value.meta.original` as the place to put the original unmodified value(s) rather than using `msg.value.cyphertext`.

## Background

Previously this module could safely assume that it would be the only one adding metadata to messages, so properties like "private", "unbox", and "cyphertext" were hardcoded. With the inclusion of `addMap()`, other applications may modify message data and add metadata. This PR makes two assumptions:

- Unboxers and map functions should only add properties at `msg.value.meta`.
- Unboxers and map functions should only make modifications to `value` after backing up the original value to `msg.value.meta.original`.

## Examples

- The unboxer sets `msg.value.meta.unbox = 'secret'` rather than creating `msg.value.unbox`.
- The unboxer modifies `msg.value.content` *after* setting `msg.value.meta.original.content`.

### Before

```json
{
  "previous": "%05hDl4m7T356FfhiJ5XTK087ULHdLR4uYGpf5RNl2ws=.sha256",
  "sequence": 6213,
  "author": "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
  "timestamp": 1539794580995,
  "hash": "sha256",
  "content": {
    "type": "test",
    "text": "Hello, meta world!",
    "recps": [
      "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519"
    ]
  },
  "signature": "bTFfPHDCgpzjtc2M/d5teOlgOTnRMXa5aI1eESKD0smH952H2+g1QdIsoCI/qwh54xE27Ljaet+Q8WGczQajBg==.sig.ed25519",
  "cyphertext": "CEYfQ/7PrUHzZWlSklgWewHp/dMfVHJN3ORZFsiBxzfTFYQD+l5j9Zvd46e2JiFAFjSl+yf5o2tEoF/t/BeY6I22R2/lBXXKkzZ7PKfW6ig75QFEdzRkpuljVN/c1nPIj0SeiU1AV2wK+RCLCHemGUpWObML92xojEBGq2yEdqfryYSZRoaQxoxzf4TKfp/ienSxohFTF0ChfqOQEyvRwrUhjFFrV1sCZ0EImrl2u8SUfmM9IDz6cOwo75DKY3w97qtpuO7GrGDO2lG3BehYSA3vsaLiWdZMPietDw0WqCVMOSO36Nk=.box",
  "unbox": "AU1F31+HaxxzFzxuD4BzMZ80rvAp63BwxBgZsr+M2NRi",
  "private": true
}
```

### After 

Note that "cyphertext", "private", and "unbox" are repeated for backward-compatibility, but may be deprecated in the future.

```json
{
  "previous": "%05hDl4m7T356FfhiJ5XTK087ULHdLR4uYGpf5RNl2ws=.sha256",
  "sequence": 6213,
  "author": "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
  "timestamp": 1539794580995,
  "hash": "sha256",
  "content": {
    "type": "test",
    "text": "Hello, meta world!",
    "recps": [
      "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519"
    ]
  },
  "signature": "bTFfPHDCgpzjtc2M/d5teOlgOTnRMXa5aI1eESKD0smH952H2+g1QdIsoCI/qwh54xE27Ljaet+Q8WGczQajBg==.sig.ed25519",
  "meta": {
    "original": {
      "content": "CEYfQ/7PrUHzZWlSklgWewHp/dMfVHJN3ORZFsiBxzfTFYQD+l5j9Zvd46e2JiFAFjSl+yf5o2tEoF/t/BeY6I22R2/lBXXKkzZ7PKfW6ig75QFEdzRkpuljVN/c1nPIj0SeiU1AV2wK+RCLCHemGUpWObML92xojEBGq2yEdqfryYSZRoaQxoxzf4TKfp/ienSxohFTF0ChfqOQEyvRwrUhjFFrV1sCZ0EImrl2u8SUfmM9IDz6cOwo75DKY3w97qtpuO7GrGDO2lG3BehYSA3vsaLiWdZMPietDw0WqCVMOSO36Nk=.box"
    },
    "private": true,
    "unbox": "AU1F31+HaxxzFzxuD4BzMZ80rvAp63BwxBgZsr+M2NRi"
  },
  "cyphertext": "CEYfQ/7PrUHzZWlSklgWewHp/dMfVHJN3ORZFsiBxzfTFYQD+l5j9Zvd46e2JiFAFjSl+yf5o2tEoF/t/BeY6I22R2/lBXXKkzZ7PKfW6ig75QFEdzRkpuljVN/c1nPIj0SeiU1AV2wK+RCLCHemGUpWObML92xojEBGq2yEdqfryYSZRoaQxoxzf4TKfp/ienSxohFTF0ChfqOQEyvRwrUhjFFrV1sCZ0EImrl2u8SUfmM9IDz6cOwo75DKY3w97qtpuO7GrGDO2lG3BehYSA3vsaLiWdZMPietDw0WqCVMOSO36Nk=.box",
  "private": true,
  "unbox": "AU1F31+HaxxzFzxuD4BzMZ80rvAp63BwxBgZsr+M2NRi"
}
```

## Deprecations

This PR leaves the old property names in-place for backward-compatibility, but if this is merged they should be considered deprecated. To be specific, this PR:

- Deprecates `msg.value.private` in favor of `msg.value.meta.private`
- Deprecates `msg.value.unbox` in favor of `msg.value.meta.unbox`
- Deprecates `msg.value.cyphertext` in `msg.value.meta.original.content`

## Considerations

I think `msg.meta` *may* be more intuitive than `msg.value.meta`, but I wanted to keep this PR as close to the current code as possible. There's no rush to migrate clients from checking `msg.value.private` to `msg.value.meta.private`, but migration would be difficult for functions that accept `msg.value` which have no way to access `msg.meta`. Simply put, I'd like to make sure that this PR is both backward-compatible *and* provides an easy migration path.

## But does it work?

I think so! It passes `npm test` for this module and Scuttlebot, and I can verify that replication is working as expected with *both* the example map in the README and [`ssb-blob-content@0.3.0`](https://gitlab.com/christianbundy/ssb-blob-content). All together now!

```json
{
  "key": "%ObfT8f9avwIudyBDgcYeDpY0c586/yYyQoYU6uIA6LE=.sha256",
  "value": {
    "previous": "%GdO8Ta85VyKINspiht2zhM1wK4ylEKN9GyUc/o8TLgE=.sha256",
    "sequence": 6138,
    "author": "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
    "timestamp": 1539704115399,
    "hash": "sha256",
    "content": {
      "type": "post",
      "text": "fizzBuzz!"
    },
    "signature": "A2lPO54HzqM/5QHT29sY2Qpx4zLUz/Q7c3eiTy2C/MSXhXJl+07x+hBKGkdXqBZKrhndoeP7zsKAMDHaRzQuDg==.sig.ed25519",
    "meta": {
      "original": {
        "content": "0EyarWwCHimzaArmyrQ9C4mN47gyfgQtEW7ocdf+lrMQ4/IGnFPOqfV/pjlp9dP4rOsOHolV6W4Yeb7UKmJVqinJUTPcQcMIvJYJyQgy9tjND41rSqL63QW3DTAuJzQike+3AgdcGWfVha/9lHS+kMSjVtxvqAXeiqFtbG8ZexikVN7YcmX/cHPpupyezebGq4rn8ePtn88I1ZjGB8E+DvELr07tjDiQCpSmPviqvpHuR1MzdH3iMQvEFWQUmuH7wOeqQJR7+2R/orZ7+VkRh5kVXpG5g6bAfrTQFYPeOUK7ukB1y31hxrAsecNv6XdSiky1s4nJ6NOFS3wQi7WE9cwy1vd3+ASvkOSovhQ/0KHJNmYw5V2fGA==.box"
      },
      "private": true,
      "unbox": "AsMTLUhjSf4Ad/dfz7vUJ3dKc9Xbgm9GcvkwcHX/DGM0",
      "fizz": true,
      "buzz": true
    },
    "cyphertext": "0EyarWwCHimzaArmyrQ9C4mN47gyfgQtEW7ocdf+lrMQ4/IGnFPOqfV/pjlp9dP4rOsOHolV6W4Yeb7UKmJVqinJUTPcQcMIvJYJyQgy9tjND41rSqL63QW3DTAuJzQike+3AgdcGWfVha/9lHS+kMSjVtxvqAXeiqFtbG8ZexikVN7YcmX/cHPpupyezebGq4rn8ePtn88I1ZjGB8E+DvELr07tjDiQCpSmPviqvpHuR1MzdH3iMQvEFWQUmuH7wOeqQJR7+2R/orZ7+VkRh5kVXpG5g6bAfrTQFYPeOUK7ukB1y31hxrAsecNv6XdSiky1s4nJ6NOFS3wQi7WE9cwy1vd3+ASvkOSovhQ/0KHJNmYw5V2fGA==.box",
    "private": true,
    "unbox": "AsMTLUhjSf4Ad/dfz7vUJ3dKc9Xbgm9GcvkwcHX/DGM0"
  }
}
```


christianbundy:
I've found a weird bug that I can't wrap my head around, but I'm hoping it's simple for someone else: `sbot get` is showing the original post (from the database) but `sbot get --private` is showing the modifications correctly. This isn't a private message, but I'm unclear on when clients would call one versus the other.

```console
$ sbot get --id %r2uawYMkQYsLIDGMFiRvFNGhYEgBWU7VwdxMgikeuPg=.sha256
{
  "previous": "%H1nMirEB3SNDdMN7H4iRb4IAB4lbWF+jlcs72MYYUxQ=.sha256",
  "sequence": 6216,
  "author": "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
  "timestamp": 1539797958229,
  "hash": "sha256",
  "content": {
    "type": "blob",
    "blob": "&KOTNrO62vdHQ+xVu/6SFUUSLLjFv8PGTj59cJYxybSA=.sha256"
  },
  "signature": "FjdQ5Mz6AdDa0Rm6uZOVolqj9HAZtY9npAe7V+m9e4zZyaNuFScd2jPOVpl3pdUcSvjAQLQZgP96/2Aw29dCDA==.sig.ed25519"
}
$ sbot get --id %r2uawYMkQYsLIDGMFiRvFNGhYEgBWU7VwdxMgikeuPg=.sha256 --private
{
  "previous": "%hAoOUiKm6yZ15EP7wWCSLoo7FCbSD2MeOfJpthqdNuc=.sha256",
  "sequence": 6076,
  "author": "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
  "timestamp": 1539558589114,
  "hash": "sha256",
  "content": {
    "type": "test",
    "content": "I'm shocked that the CLI works..."
  },
  "signature": "YfRpeTRzTKLGcJHFdiEH2lOGkSiKU9yAcbu+YLNewViE6yIMoF43IkguZUG4AfhG7zQnxsqomRn/jb4gqLcaDw==.sig.ed25519",
  "meta": {
    "original": {
      "content": {
        "type": "blob",
        "blob": "&jSTeGBZcdEpPOFMDzqmn8sMF7jlbe8JZOqdgeH2iWMI=.sha256"
      }
    },
    "blobContent": true
  }
}
```

christianbundy:
Captain's log: this now passes `npm test` for this module, for Scuttlebot (yes, [with an `addMap()`](https://github.com/ssbc/scuttlebot/pull/560)), and also seems to be replicating flawlessly with real-world usage.

dominictarr:
I think @ahdinosaur's patchwork message reveal is the only thing that's actually using `unbox` etc so I think we can just coordinate this release with him and easily handle the deprecation. hmm, do the clients otherwise actually use `private: true` ? (hmm, think I did that in `patchless`, but nothing there touches cyphertext or unbox)

christianbundy:
> I think @ahdinosaur's patchwork message reveal is the only thing that's actually using `unbox` etc

Just to clarify, does this mean that most clients are using `db.get({ id })` without `{ private }`? If so, I think we're unboxing and immediately reboxing every message from the database, which would definitely be worth deprecating. If backward-incompatibility is a viable option I'd be happy to drop support for `{ private }` and instead require `{ original: true }` for the few queries that need to be cryptographically verifiable. Does that sound like it would work?

christianbundy:
## Before

- **Default:** Rebox message.
- **`{ private: true }`:** Unbox message.
- **`{ private: true, unbox: 'secret' }`:** Unbox message with secret key.

## After

- **Default:** Unbox message.
- **`{ original: true }`:** Restore original message with `msg.value.meta.original`.
- **`{ unbox: 'secret' }`:** Unbox message with secret key.

## Performance

I mea

christianbundy:
The most recent change was abstracting our `metaBackup()`, which is a general abstraction over the old `msg.value.cyphertext = msg.value.content` pattern. Instead, it creates (or inserts into) `msg.value.meta.original`, so that we can just snag `msg.value.meta.original.content` directly without any fuss.

I've added documentation for this as well, but please let me know if there's a better name for it.

christianbundy:
I think this is finally stable enough for another pair of eyes (again). There are three changes:

- **Patch:** The unboxer no longer requires `{ private: true, unbox: 'secret' }`, so we can set *only* the unbox key and it should work as-is.
- **Minor:** The unboxer now places metadata in `msg.value.meta` and `msg.value` (for backward-compatability). This makes it possible for other functions to safely change data or add metadata.
- **Minor:** A new function was added to `util.js` for backing up metadata properties (e.g. `msg.value.content`). This makes it easier to back up metadata without reinventing the wheel.
- **Major:** Get/streams only return original messages if you set `{ original: true }`, whereas in the past they've returned original messages unless you set `{ private: true }`. This creates a solution less specific to the private message unboxer, which avoids reboxing by default, and resolves #222.

Most likely relevant to: @dominictarr @arj03 @ahdinosaur 

Any concerns with this branch, or anything I should document better?

dominictarr:
@christianbundy sorry I didn't reply until now, but _private absolutely must not be the default_. some apis that should not see messages contents such as _createHistoryStream_ and _ooo.get_ will otherwise see private messages! when merging this feature the first time I took great care to make sure that no private information was inadvertently leaked while replicating etc

dominictarr:
Also, in earlier versions, everything returned cyphertext by default and the client had to unbox it. `private` being false by default was to be compatible with that.

dominictarr:
I know it's slightly awkward to have `private: true` also apply to mappings that have nothing to do with privacy... but unless there is a way to allow those through but not private messages, I think it would be simpler to just keep `private: true` as off by default, as it was.

util.js
```diff
@@ -44,19 +44,21 @@ exports.wait = function () {
...
-  const oldProps = ['cyphertext', 'private', 'unbox']
-
-  const metaProps = oldProps.concat('meta')
-  const original = value.meta && value.meta.original || {}
-
-  var o = {}
-  for (var key in value) {
-    if (!metaProps.includes(key))
-      o[key] = original[key] || value[key]
+  if (value.meta) {
```
dominictarr: I think here how you are mutating and before it was copying will break stuff, because I believe these objects are cached, so modifying them will change the cached value. A subsequent request, that asks for the mapped data will get it unmapped because of this. (unless it's read for disk again - which will be quite hard to debug!)

util.js
```diff
@@ -44,19 +44,21 @@ exports.wait = function () {
...
+    if (value.meta.original) {
+      Object.entries(value.meta.original).forEach(entry => {
```
dominictarr: btw, what values get stored here? isn't it just content?

util.js
dominictarr: it's also essential the order of the keys are preserved, which is why it uses a for each in the previous version.

util.js
christianbundy: Only `content`, but I couldn't detect a performance difference between hardcoding `content` and allowing *any* property of `msg.value` be backed up. For example, if you wanted to transform messages to fix impossible timestamps (e.g. asserted timestamp is in 2060 when received timestamp is in 2018) then it would be important to be able to have the original backed up to `meta.original.timestamp`.

util.js
christianbundy: Nice catch, thanks. I didn't consider that.

dominictarr: @arj03 fixed that problem the first time :wink:

indexes/clock.js
```diff
@@ -38,9 +38,9 @@ module.exports = function (db, opts) {
           sync: false === (opts && opts.sync),
           limit: limit
         }),
-        //NEVER allow private messages over history stream.
-        //createHistoryStream is used for legacy replication.
-        u.Format(keys, values, false)
+        // NEVER allow private messages over history stream.
+        // createHistoryStream is used for legacy replication.
+        u.Format(keys, values, true)
```
christianbundy: @dominictarr This doesn't change the behavior of `createHistoryStream`, which continues to output *only* the original message, but I don't mind going back to checking for `opts.private` if that's the most ergonomic.

christianbundy:
No problem, I'll work on reversing the implementation back to the previous default, so that the transformations are only returned when `{ private: true }` is set.

indexes/clock.js
```diff
@@ -55,7 +55,7 @@ module.exports = function (db, opts) {
       opts.keys = false
       opts.values = true
 
-      return pull(index.read(opts), u.Format(keys, values, opts.private === true))
+      return pull(index.read(opts), u.Format(keys, values, opts.private))
```
dominictarr: this should be `opts.private === true` so that it's only if exactly `true` not just truthy, which is more likely to happen by accident.

christianbundy: Thanks, I've resolved this in `u.Format` instead so that we only have to check it once. See: https://github.com/ssbc/secure-scuttlebutt/pull/228/commits/4f883fce0c3e9f7d3d39a36609f893be9f7d5d43

dominictarr:
oh, I thought you'd removed all the `opts.private === true` and that made `private: 1` and other truthy values work, but then I realized that formatStream checks `isPrivate === true` so it's okay. merging!

dominictarr:
merged into 18.6.0! (I'm also about to use this in private-groups, so thanks!)

dominictarr:
merged into 18.6.0

PR #231, opened 2018-10-23
Remove spacing from encoded JSON

christianbundy:
This is where my SSB knowledge starts to get hazy, but I *think* this means that we'll end up storing a smaller JSON payload in our flume log, which [quick perf test](https://jsperf.com/ssb-json-parse-json-stringify) suggests makes this about 10% faster:

```js
codec.decode(codec.encode(obj))
```

This passes `npm test` for this repo and Scuttlebot, and making this change to bench-ssb's  `02-json.js` results in 6% less time being spent stringifying JSON. I've also removed the `flumecodec` packages since we haven't been using it, but once #227 or #228 are merged then we *could* just use the vanilla codec from the package. Makes no difference to me.

@dominictarr @arj03 Is this going to shoot us in the foot for any reason?


dominictarr:
> Is this going to shoot us in the foot for any reason?
yes, maybe. the signing protocol still uses spaced json, although the network doesn't. It would probably have made more sense to use the same thing everywhere, any way too late there.

please run some of the `bench-ssb` benchmarks, 10% is really a lot - i want to know if it makes that much difference when it's used with a whole app

arj03:
This will propagate to the UI right, so in that regard it is a change. Also agree that maybe we are a bit too late with this. I seems as if the new formats might be closer that I originally though.

I've been banging at the performance part for quite a while now. Thought I could get the json performance better using turbo-json. Turns out that the messages are a big mess, especially about and post, which negates most of the performance gains. At first it was quite a bit faster, then I ran it through the whole log with validate to make sure we generate exactly the same output and after that the performance was not good enough to justify the extra complexity :( That was VERY frustrating.

I've been changing tactics and are trying to fix the performance of validate. That is a very big overhead. See https://github.com/ssbc/secure-scuttlebutt/pull/223. I'll have another go at it tonight and see if it can be done better.

AljoschaMeyer:
Fyi, the rust implementation for legacy messages is close to done, there'll be js bindings soon.

@arj03, what kind of functions would you need from rust? The set of function I was currently planning:

- take some bytes encoding a message, parse them into a js object
- take a js object, encode it into the whitespace-less transport json encoding
  - this still depends on v8 object entry iteration order
- take a js object, encode it into the signing encoding
  - this still depends on v8 object entry iteration order
- take some bytes, check whether they encode a valid message, and compute the corresponding hash, length, and signing encoding (this gets to use a more efficient data structure than js objects)
- take some bytes and compute the corresponding hash and length, skipping any validation

christianbundy:
> This will propagate to the UI right, so in that regard it is a change. Also agree that maybe we are a bit too late with this. I seems as if the new formats might be closer that I originally though.

I'm under the impression that this *only* changes the IO with flumelog-offset, is that wrong? I think I've got a blind spot on how this would propagate to the UI or make any other changes, as I *think* this is just how we read/write from the flumelog.

## Benchmarks

I ran each of these benchmarks 10 times on each branch (note to self: automate bench-ssb diffs!) and averaged them:

### `master`

name        | seconds | percent (haha)
------------|---------| ---
json        | 0.2327| 100%
minimal     | 13.6499 | 100%
legacy      | 23.4814 | 100%
random-read | 2.013 | 100%

### `faster-json`

name        | seconds | percent
------------|---------|---
json        |  0.2218| 95.32%
minimal     | 13.6777| 100.2%
legacy      | 23.3454| 99.42%
random-read |  1.9508 | 96.91%


Looks like random-read is about 3% faster?

## Code

Code that made my life easier while benchmarking:

```sh
# run a benchmark multiple times, append final value to file
: > ssb-read.txt; for run in {1..10}; do echo $run; node 06-ssb-read.js | tail -q -n 1 | awk '{ print $1, $5 }' >> ssb-read.txt; done
# take an average from the file we just created
awk '{ total += $2; count++ } END { print total/count }' < ssb-read.txt
```

arj03:
> @arj03, what kind of functions would you need from rust? The set of function I was currently planning:

Bear with me, I'm really tired, but I guess you have something that does the same thing as ssb-validate does? 

Secondly there is the JSON.stringify and JSON.parse code. Have you tested your rust implementation against node in this regard? Would be quite interesting to see :-) What I did when I was benchmarking an alternative json implementation was to read messages from the database to test. I have something like 570k locally so it was a pretty good test case :-)

AljoschaMeyer:
@arj03 Yup, there are (de)serialization functions and (soon) length-check, hash computation, validity of the `"type" in the content, and signature verification. All the other checks are part of the parser (which is hardcoded for the precise metadata format rather then parsing into generic json first).

So the deserialization is an efficient combination of JSON.parse and ssb-validate. And the deserialization will also be able to directly serialize into js objects, without any intermediate data structure. So I'm pretty optimistic that this will improve performance over the current js codebase.

arj03:
Great. Let me know when there is something we can try and hook up :)

AljoschaMeyer:
@arj03 If you could wish for any kind of js API, what would you want?

arj03:
Well, lets start with:
 - json = deserialize(buffer)
 - json = deserializeAndValidate(buffer, previous json object)
 - bool = validate(json object, previous json object)
 - buffer = serialize(json object)

Not 100% sure that is all we need, but it would be a good start :)

AljoschaMeyer:
Other stuff that might be useful:

- would you benefit from splitting syntactic verification and signature verification?
- what about length-computation and hash computation? How often and at which points does sbot compute those? Would something like `(json, hash, length) = deseriallizeAndValidateAndHashAndLength(buffer, previous)` be helpful?
- should validation really take the previous json object, or should it just be the hash of the previous message?

Do you need any of these?

arj03:
1) If you can split them, then sure, just as long as there is one that does both
2) Right, (json, hash, length) = deseriallizeAndValidateAndHashAndLength(buffer, previous) would be better than just returning json
3) it needs the previous object as it needs sequence as well as hash.

arj03:
Also, if you can, split so we hash and length as seperate functions as well. Might be useful.

christianbundy:
> I'm under the impression that this *only* changes the IO with flumelog-offset, is that wrong?

Following up on this, I've been running this for about a week and all of my messages seem to be propagating fine. The only changes this branch makes are:

- no `toString()` on JSON that should already be a string
- slightly faster codec because we aren't storing the message with spacing
- slightly smaller storage cost because we can strip the spacing
  - small posts like `%ityWj4brumwpUL1JW8T7rgStPFQDAwAiEbI2msiQOIA=.sha256` are 20% smaller 
  - big posts like `%nbVzQ1pX/3PwDhu4h0j1OQaBiIbCqJ0KYB9C/f5grzU=.sha256` are 8% smaller

@arj03 @dominictarr
Are there any other issues I should be looking out for?

arj03:
Little low on time. Would like to test this though. But have you checked on patchbay that the raw messages are still pretty printed?

christianbundy:
No problem at all, I appreciate the suggestions of which things should be tested!

Raw messages in Patchbay are still pretty-printed the same as before, I think this PR *only* changes how messages are stored at `~/.ssb/flume/*.json`. I think it's the equivalent of [this line](https://github.com/ssbc/bench-ssb/blob/e269925c2be8baa1ba90e2db40b099194ffbcccf/00-generate.js#L41) in bench-ssb.

christianbundy:
Any objections to merging this? It's smaller, simpler, and faster, and I don't think (?) there are any downsides.

arj03:
Testing this now.

arj03:
No problem so far. Great :+1:

PR #235, opened 2018-10-26
Remove unused `test/group.js`

christianbundy:
Was this file being used for anything? I couldn't find any references with `git grep`.


cryptix:
See the test script in the pkg.json. It uses `test/*.js`

christianbundy:
I saw that it's being run, but I'm confused at what it's testing. I'd be happy to document it if this is an important test, but it doesn't look like it's even importing any of our code.

christianbundy:
It looks like this was added as a utility function for [an old test](https://github.com/ssbc/secure-scuttlebutt/blob/4d5d5bfb7096e8c21dedd1280f25bb94d5cf5231/test/net-replicate.js), but I don't think we're using this at all anymore. Maybe I'm looking in the wrong place?

PR #238, opened 2018-10-30
Key undefined

### LINKED ISSUE #237: `key` is undefined
See: %h4J9aYMcmT68dtiZNAP6pS0LoNnSRXzKuNqXMTDrnDs=.sha256

```
~/.npm-global/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:41
            msg.unbox = key.toString('base64')
                            ^
TypeError: Cannot read property 'toString' of undefined
    at unbox (~/.npm-global/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:41:29)
    at Object.decode (~/.npm-global/lib/node_modules/scuttlebot/node_modules/secure-scuttlebutt/minimal.js:88:14)
    at ~/.npm-global/lib/node_modules/scuttlebot/node_modules/flumelog-offset/inject.js:45:22
    at ~/.npm-global/lib/node_modules/scuttlebot/node_modules/flumelog-offset/frame/recoverable.js:48:11
    at ~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:74:11
    at get (~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:29:7)
    at next (~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:53:7)
    at ~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:63:25
    at get (~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:29:7)
    at next (~/.npm-global/lib/node_modules/scuttlebot/node_modules/aligned-block-file/blocks.js:53:7)
```

christianbundy:
Resolves #237 

This diff is just an indentation fix plus:

```diff
-msg.unbox = key.toString('base64')
+if (key) {
+  msg.unbox = key.toString('base64')
+}
```

I'm unclear on *how* this could be happening, but it seems that the following must be true to trigger this error:

- `isFunction(unbox) === false`
- `key == null`
- `unbox.key == null`

@dominictarr @ahdinosaur Is this fine, or should we throw an error if we can't set `msg.value.unbox`?

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -22,12 +22,11 @@ function unbox(data, unboxers, key) {
     for(var i = 0;i < unboxers.length;i++) {
       var unboxer = unboxers[i]
 
-      if (key) {
-        plaintext = unboxer.value(data.value.content, key)
-      } else if (isFunction(unboxer)) {
+      if (isFunction(unboxer))
         plaintext = unboxer(data.value.content, data.value)
-      } else if (!key && unboxer.key) {
-        key = unboxer.key(data.value.content, data.value)
+      else {
+        if (!key) key = unboxer.key(data.value.content, data.value)
+        if (key) plaintext = unboxer.value(data.value.content, key)
       }
 
       if(plaintext) {
@@ -235,4 +234,15 @@ module.exports = function (dirname, keys, opts) {
 
   return db
 }
+
+
+
+
+
+
+
+
+
+
+
 
--- a/minimal.js
+++ b/minimal.js
@@ -21,15 +21,14 @@ function unbox(data, unboxers, key) {
   if(data && isString(data.value.content)) {
     for(var i = 0;i < unboxers.length;i++) {
       var unboxer = unboxers[i]
-      if(isFunction(unboxer)) {
+
+      if (key) {
+        plaintext = unboxer.value(data.value.content, key)
+      } else if (isFunction(unboxer)) {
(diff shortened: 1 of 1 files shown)


dominictarr:
this could happen if someone has an old unboxer that doesn't do the key separately, this could happen.

I also see some other problems with this code - the top `if(isFunction...) else ... ` should all be inside `if(key) {...}` because if a key is provided, it should be used for decryption explicitly.

minimal.js
```diff
@@ -20,16 +20,15 @@ function unbox(data, unboxers, key) {
   var plaintext
   if(data && isString(data.value.content)) {
     for(var i = 0;i < unboxers.length;i++) {
-      var unbox = unboxers[i]
-      if(isFunction(unbox)) {
-        plaintext = unbox(data.value.content, data.value)
+      var unboxer = unboxers[i]
+
+      if (key) {
+        plaintext = unboxer.value(data.value.content, key)
```
christianbundy: @dominictarr Like this?

dominictarr: almost...

christianbundy:
Oh, that makes sense. Thanks!

PR #239, opened 2018-10-31
Wait 100ms before writing instead of 10ms

arj03:
I was testing 05-ssb-legacy write to see where the time goes. Comparing with 01 validate I get validate: 53s, write + validate (05) 125s. Seems rather high for 42mb data. Looking around I found the async-write function that defaults to delay write 100 ms, but for some reason it set to 10ms in secure-scuttlebutt. Changing it to 100 ms makes a large difference. 15 seconds!

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -140,7 +140,7 @@ module.exports = function (dirname, keys, opts) {
     return state.queue.length > 1000
   }, function isEmpty (_state) {
     return !state.queue.length
-  }, 10)
+  }, 100)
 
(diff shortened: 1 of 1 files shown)


arj03:
Also I tried with 200 and 50, didn't make much diff. Think 100 should be a better default.

dominictarr:
looks good, please merge.

(am a little bit worried about fixed numbers like this, would be great if it could figure out the best number on it's own... too complicated though)

PR #245, opened 2018-12-29
also pass value to unboxValue

dominictarr:
needed by private-groups,

so that previous is accessable, for instance. https://github.com/dominictarr/group-box needs an external nonce to decrypt (which is supplied as the previous hash)

### DIFF
--- a/minimal.js
+++ b/minimal.js
@@ -70,7 +70,8 @@ function isString (s) {
 }
 
 module.exports = function (dirname, keys, opts) {
-  var hmacKey = opts && opts.caps && opts.caps.sign
+  var caps = opts && opts.caps || {}
+  var hmacKey = caps.sign
 
   var mainUnboxer = {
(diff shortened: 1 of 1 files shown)


dominictarr:
we have addUnboxer also needed a way for plugins to encrypt messages. I now realize that we could/should have just used hooks for handling unboxing too. hmm.

christianbundy:
Wasn't sure after your last comment: is this PR still good to merge from your side? It seems like the changes are good to me:

- Pass more data to the unboxer, if it wants it.
- Move `box()` into its own function.

dominictarr:
I've changed my mind 180 degrees on the last comment. to get this working we added a new method, `sbot.addUnboxer(fn)` instead of using `sbot.unbox.hook(fn)` hooks a weird and make implicit deps, and no one understands them, but every method gets hooked. That's just modular madness. better to be really explicit about what the extention points are. I now think it would be much better to rewrite to make these points of extention explicit, like `addUnboxer`

PR #247, opened 2019-01-04
remove legacy level support

dominictarr:
it's pretty old now, but before we introduced flumedb, everything was in a leveldb.
as recently discovered: https://github.com/ssbc/ssb-db/pull/246 there is some problems with that. I don't think it's worth figuring out though, since most people are running the new code, and if you really arn't, then you can still resync. Also this clears out quite a bit of cruft and is a good step towards https://github.com/ssbc/ssb-server/issues/602

This is a breaking change, so should probably not bundle it into what happens for https://github.com/ssbc/ssb-server/issues/602


arj03:
Very much in favor of removing things like this.

index.js
```diff
@@ -17,21 +17,14 @@ function errorCB (err) {
   if (err) throw err
 }
 
-module.exports = function (_db, opts, keys, path) {
+module.exports = function (_, opts, keys, path) {
+  //_ was legacy db. removed that, but for backwards compatibilty reasons do not change interface
   path = path || _db.location
```
staltz: How about also removing or changing this reference to `_db`?

dominictarr: fixed.

dominictarr:
this PR is also now the ssb-db end of https://github.com/ssbc/ssb-server/pull/610

With this, you can use ssb-db directly as a plugin.
So you don't need to `require('ssb-server')` at all, instead just do: `SecretStack().use(require('ssb-db').use(...)`.

dominictarr:
note: this is a breaking change.

staltz:
> note: this is a breaking change.

Just remember to release the next version as a bump on the major number. Fine by me to merge

christianbundy:
I noticed that this introduces a new `api.md` file, should the API section be removed from the readme so we don't have duplicate content?

dominictarr:
yes probably the documentation should be consolidated so there is one best documentation.
a restriction though is that api.md is also machine-read, it's loaded by https://www.npmjs.com/package/mdmanifest and used to produce the mainfest for this plugin. (actually all the sbot plugins had that)

staltz:
I never liked this mdmanifest idea, but particularly on mobile it's a tricky thing to bypass because it uses fs.readFileSync inside the node_modules, in runtime, which doesn't play well with noderify, and I had to [hack a way to replace these readFileSync calls](https://gitlab.com/staltz/manyverse/blob/a5e097a99dc36ac6150665394a3a1b8c4211cc43/tools/patch-rnnodeapp-scuttlebot) with the actual file content in build time.

PR #250, opened 2019-01-22
add osenv to package.json dependencies

staltz:
`osenv` is required in index.js https://github.com/ssbc/ssb-db/blob/e97198694e61ea96150c948198f516f4f2223005/index.js#L5 but was missing from package.json


mixmix:
likewise this is in my package.json

PR #251, opened 2019-01-22
add mkdirp to package.json dependencies

staltz:
`mkdirp` is required in index.js https://github.com/ssbc/ssb-db/blob/e97198694e61ea96150c948198f516f4f2223005/index.js#L6 but was missing from package.json.


mixmix:
this is odd .. I can see it in the master branch ...

PR #252, opened 2019-01-22
add rimraf to package.json dependencies

staltz:
`rimraf` is required in index.js https://github.com/ssbc/ssb-db/blob/e97198694e61ea96150c948198f516f4f2223005/index.js#L7 but was missing from package.json.


arj03:
It is in devDependencies @staltz, seems like it needs to be moved instead.

PR #254, opened 2019-03-05
get ({id: id, meta: true}) to get {key,value,timestamp} format message

dominictarr:
it is kinda annoying that lots of things use a `{key,value,timestamp}` pattern but `sbot.get` doesn't do that.
this tends to mean that sometimes messages don't have received times... this adds a option to keep that.

`meta` is short for metadata. using this name is consistent with other uses for example `links` which uses takes a `meta` option (although that one defaults to true)

Making meta default to true here is probably a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,18 @@ ssb-db works.
 
 The following methods all apply to a `ssb-db` instance
 
+### SSBdb#get (id | seq | opts, cb)
+
+Get an ssb message. If `id` is a message id, the message is returned.
+If seq is provided, the message at that offset in the underlying flumelog
+is returned. If opts is passed, the message id is taken from either
+`opts.id` or `opts.key`. If `opts.private = true` the message will be decrypted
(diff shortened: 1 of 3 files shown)


christianbundy:
I dig it. Is this behavior already documented somewhere else, or do we need docs? The tests LGTM, thanks!

Happy0:
Nice, @dominictarr - this is actually something that I ran into in a recent change I made to ssb-chess where I was surprising I couldn't do `msg.key` to get the message ID in a helper function I had passed the message to.

I like that this is implemented in a backward compatible way :)

dominictarr:
@christianbundy now that you mention it... ssb-db's documentation needs a complete overhaul. There is stuff that's completely wrong now, or is left from before scuttlebot even existed I think. get isn't even documented!

dominictarr:
I didn't have the time to do that overhaul, but I wrote documentation for get at least, including this feature.

merged into 19.1.0

PR #255, opened 2019-03-12
documentation drive: document most important methods, list still undocumented methods, shift deprecated methods to deprecated section

dominictarr:
The documentation was previously _very bad_ this improves it significantly.

rate the current state of the documentation out of 10 - 2 (now much improved)
is it clear what this is used for? - needs more guidance around how to create ssb plugins, and setting up minimal ssb set ups.
is there api documentation? - Yes
read code, do you see any undocumented methods? - quite a few, but these are explicitly listed.
for a method is there any undocumented options? - probably, yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ssb-db
 
-A database of unforgeable append-only feeds, optimized for efficient replication for peer to peer protocols.
+secret-stack plugin which provides storing of valid ssb messages in an append-only log.
 
 ## What does it do?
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -119,13 +119,11 @@ module.exports = {
 
       since                    : since,
 
-      getPublicKey             : ssb.getPublicKey,
       latest                   : ssb.latest,
       getLatest                : valid.async(ssb.getLatest, 'feedId'),
       latestSequence           : valid.async(ssb.latestSequence, 'feedId'),
       createFeed               : ssb.createFeed,
       whoami                   : function () { return { id: feed.id } },
-      query                    : ssb.query,
```
arj03: Are these not used anywhere?

arj03:
Wow this is a massive improvement! +1

README.md
```diff
@@ -175,21 +172,17 @@ database does not have the private key:
...
 
-Create an ssb-db database at the given path, returns an instance.
+Adds ssb-db persistence to a secret-stack setup.
+Without other plugins, this instance will not have replication
+or querying. Loading ssb-db directly is useful for testing,
+but it's recommended to instead start from a plugin bundle like [ssb-server](https://github.com/ssbc/ssb-server)
 
-### require('ssb-db')(db, opts)
+Because of legacy reasons, all the ssb-db methods are mounted on the top level object,
+so it's `sbot.get` instead of `sbot.db.get` as it would be with all the other `ssb-*` plugins.
```
jedahan: Do methods also bind to `sbot.db`, or only to the top level object? Should I open an issue to have it bind to both, in the eventual hope we can migrate to the non-legacy binding?

dominictarr: they bind to the top level. yeah I was thinking we should move them to sbot.db and then have a thing that makes it backwards compatible (but print a deprecation message)

index.js
dominictarr: query isn't even set! latest,getLatest,latestSequence are used in some tests. createFeed is mostly used in tests. this is the worst case because changing the tests isn't directly useful, but it would allow us to remove a bunch of code.

PR #257, opened 2019-03-14
documentation drive: document what was undocumented

dominictarr:
this PR finishes documenting all the undocumented methods.

addUnboxer and unbox definitely need more work.
also, I noticed there are several methods (about getting the latest message) that do not seem to be in use anymore, they should probably be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,25 +356,126 @@ which is why it starts with `_`.
 
 see ***undocumented*** creating a [secret-stack](https://github.com/ssbc/secret-stack) plugin.
 
-## Undocumented API
+### getAtSequence ([id, seq], cb(err, msg))
 
-these methods exist, and are used for something important, but are not currently documented.
+get a message at a given feed `id` with given `sequence`.
+calls back a message or an error, takes a two element array
+with a feed `id` as the first element, and `sequence` as second element.
 
-* getAtSequence     - get message at a {id,sequence}
-* getVectorClock    - get the latest sequence for every feed held in log
-* latest            - get latest `{id,sequence,ts}` for given feed, id is the message hash.
-* latestSequence    - get latest sequence number for given feed
-* getLatest         - get the latest message for a given feed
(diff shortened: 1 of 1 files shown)


dominictarr:
Also, some methods are available to the local api but not over muxrpc. sometimes this is necessary because muxrpc can't do things like observables, other times I just forgot to put it in the `api.md` each method should be explicit about wether it's exposed or not.

dominictarr:
closes: https://github.com/ssbc/ssb-db/issues/256

README.md
```diff
@@ -356,26 +356,127 @@ which is why it starts with `_`.
...
-* status            - get status information about index building
-* version           - current ssb-db version
-* queue             - append a message to write queue (without writing immediately)
-* post              - get called back when a message is appended
-* since             - observable of the current offset of underlying flumelog.
-* addUnboxer        - add a handler for unboxing (decrypting) messages
-* box               - encrypt a message content
+### getAtSequence ([id, seq], cb(err, msg))
+
+get a message at a given feed `id` with given `sequence`.
```
arj03: maybe: "for a given feed id" instead of "at a given feed" or is my english bad? :-)

README.md
```diff
@@ -356,26 +356,127 @@ which is why it starts with `_`.
...
+calls back a message or an error, takes a two element array
+with a feed `id` as the first element, and `sequence` as second element.
+
+needed for [ssb-ebt replication](https://github.com/ssbc/ssb-ebt)
+
+### getVectorClock (cb)
+
+load a map id to latest sequence (`{<id>: <seq>,...}`) for every feed in the database.
```
arj03: load a map of

arj03:
Added minor wording suggestions, looks good otherwise!

README.md
dominictarr: no your english is better than mine

dominictarr:
merged into 19.1.1

PR #259, opened 2019-04-23
Add option not to delete tmp based .ssb directory if a directory of the configured name already exists.

Happy0:
I'm currently using https://github.com/ssbc/scuttle-testbot to test something. It would be useful for a test case I'm working on to be able to retain the old `tmp` directory.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,12 @@ module.exports = {
     if(opts.temp) {
       var name = isString(opts.temp) ? opts.temp : ''+Date.now()
       opts.path = path.join(osenv.tmpdir(), name)
-      rimraf.sync(opts.path)
+
+      if (!opts.startUclean) {
+        // The 'startUnclean' option allows the caller to not remove the old temp directory with the given name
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -33,7 +33,12 @@ module.exports = {
     if(opts.temp) {
       var name = isString(opts.temp) ? opts.temp : ''+Date.now()
       opts.path = path.join(osenv.tmpdir(), name)
-      rimraf.sync(opts.path)
+
+      if (!opts.startUclean) {
```
christianbundy: I think there may be a typo here. :wink: 

Also, if possible I think it would be nice to specify that we're looking for `true` specifically.

```suggestion
      if (opts.startUnclean !== true) {
```

Happy0:
Thanks @christianbundy - can do :) (fix the bug and add a unit test.)

christianbundy:
Anything I can do to help move this forward? I'd love to see this merged.

PR #262, opened 2019-05-20
Add `del()` method to delete messages by key

christianbundy:
Woo! This adds deletion to ssb-db, which overwrites deleted content with null bytes.

When reading deleted content there are two behaviors:

- If you stream deleted content it's filtered out.
- If you `get()` a deleted key it returns an error object with `err.code === 'flumelog:deleted'`.

This replaces flumeview-hashtable with flumeview-level, as explained in the commit message for e904acb:

>The benchmarks for these look about the same but flumeview-level
supports deletion without any problems. Flumeview-hashtable supports
deletion but requires a full view rebuild each time we delete a message,
which is less than  ideal.

Anyone strongly opposed to this change?


arj03:
My emojies where for initial excitement :), will give it a proper look code wise and test it out soon.

staltz:
Superficially this looks correct, thanks for the effort!

Now, about the deletion of the message itself, what does this mean for sig-chain verification? Wouldn't verification be broken?

christianbundy:
@staltz 

Yeah, this is focused on the "I've downloaded messages I don't want on my computer, I need to make them go away" and the main use-cases is to delete content from people you've already blocked. You *could* use this if you've accidentally forked your feed and want to remove the offending message from your database, but the focus is on full feed deletion.

Full context here: https://github.com/ssbc/ssb-server/issues/651

arj03:
This seems to be working pretty well!

I ran a few of the tests in bench-ssb and it appears to be about the same speed. Then I created a small script based on your test and deleted a msg. Streaming then seamlessly leaves out the deleted message, but I can still get it based on the key, just as in the test :+1:

arj03:
Also the code change is pretty minimal, as most of the heavy lifting is already merged, so overall looks good to me.

christianbundy:
@arj03 Thanks for the review!

> Streaming then seamlessly leaves out the deleted message, but I can still get it based on the key, just as in the test

Hmm, could you confirm that you *can* still get the message based on the key? It should callback an error if you try to `get()` a deleted key:

https://github.com/ssbc/ssb-db/blob/d706b40c881817fced02218bb6961c3e74809971/test/log.js#L159-L163

arj03:
Sorry @christianbundy. I wasn't expression myself clearly. I meant to say that it works, in that if I try to get the message by key I will get an error and I can see the special attribute you added to it. So all good :-)

christianbundy:
Thanks! I've just pushed a bit more documentation to `API.md`, I'll merge and publish now.

PR #266, opened 2019-09-19
Add support for deleting entire feed at once

christianbundy:
This is what we're doing manually in Patchwork, except for in Patchwork we're also keeping track of messages that are currently in the process of being deleted because it's common to call `del()` on the same message multiple times. I'm unsure whether we should:

- Add some code to support race conditions so `del()` can be called lots without errors.
- Leave as-is and probably have scenarios where `del()` called repeatedly throws errors.

Relevant Patchwork code: https://github.com/ssbc/patchwork/blob/a7dd2056bf17f0986cf3f8e4445fac3e024fb4ef/lib/main-window.js#L81-L121

### DIFF
--- a/api.md
+++ b/api.md
@@ -322,9 +322,10 @@ Outputs information in the following form:
 
 ⚠ This could break your feed. Please don't run this unless you understand it.
 
-Delete a message by its message or feed key. This only deletes the message from
-your local database, not the network, and could have unintended consequences if
-you try to delete a single message in the middle of a feed.
+Delete a message by its message key or a whole feed by its key. This only
+deletes the message from your local database, not the network, and could have
+unintended consequences if you try to delete a single message in the middle of
+a feed.
 
 The intended use-case is to delete all messages from a given feed *or* deleting
 a single message from the tip of your feed if you're completely confident that
--- a/api.md
+++ b/api.md
@@ -322,9 +322,9 @@ Outputs information in the following form:
 
 ⚠ This could break your feed. Please don't run this unless you understand it.
 
-Delete a message by its message key. This only deletes the message from your
-local database, not the network, and could have unintended consequences if you
-try to delete a single message in the middle of a feed.
+Delete a message by its message or feed key. This only deletes the message from
+your local database, not the network, and could have unintended consequences if
+you try to delete a single message in the middle of a feed.
 
 The intended use-case is to delete all messages from a given feed *or* deleting
 a single message from the tip of your feed if you're completely confident that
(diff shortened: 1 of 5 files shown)


api.md
```diff
@@ -322,9 +322,9 @@ Outputs information in the following form:
 
 ⚠ This could break your feed. Please don't run this unless you understand it.
 
-Delete a message by its message key. This only deletes the message from your
-local database, not the network, and could have unintended consequences if you
-try to delete a single message in the middle of a feed.
+Delete a message by its message or feed key. This only deletes the message from
```
arj03: This should probably be clarified to be something like: 

Delete a message by its message or feed key

=>

Delete a message by its message key or a whole feed by its key.

create.js
```diff
@@ -34,7 +34,22 @@ module.exports = function (path, opts, keys) {
...
+    pull(
+      db.createUserStream({ id: feed }),
+      pull.asyncMap((msg, cb) => {
+        const key = msg.key
+
+        deleteMessage(key, (err) => {
+          cb(err, key)
+        })
+      }),
+      pull.filter(),
```
arj03: Why is this needed?

arj03:
Added a few comments, I havn't tested all the delete stuff, but I'm actually keen on testing it out soon :)

staltz:
Looks like a small and neat PR, would also be a handy API to have. I also echo arj's comment about that pull.filter(), but otherwise this looks good to me.

create.js
christianbundy: It was added because sometimes we were getting `null`, but that could've been a Patchwork quirk. I'll try removing it and see what happens.

staltz: Could also just put a comment `// removes nulls`

api.md
christianbundy: Good idea, thanks!

christianbundy:
Should be ready for review now, thanks again!

arj03:
@christianbundy was testing this. It doesn't work if you connect to the sbot and try to delete because of this line:

https://github.com/ssbc/ssb-db/blob/7ea0a9e2aa229c8f596e3ff10df09ed0b7837271/index.js#L124

Needs to be 'msgLink|feedId'. With that I can confirm that it works :rocket: It is quite snappy to remove a feed, nice work :)

christianbundy:
Thanks, fixed!

arj03:
Great, I would say this is good to go then!

arj03:
One more thing @christianbundy. Not directly related to this PR, but shouldn't this [line](https://github.com/ssbc/ssb-db/blob/ba52a9e718f028471cd76860e73bed5dcd816d5e/create.js#L40) have a return?

christianbundy:
...yes, yes it should. Thanks for pointing that out, I'll push a commit fixing it.

Aside: lately I've been thinking that maybe each function should only have one `return` to avoid problems like this. It would be a huge pain if functions had multiple entry points -- maybe it would be nicer if they only had one [non-`throw`] exit point as well?

arj03:
wop wop :rocket:

PR #267, opened 2019-09-23
remove mdmanifest

staltz:
mdmanifest used to be widely used in ssb libraries, but not so much nowadays. ssb-db is the only one still using, as far as I know.

In codebases like Manyverse, we have to [patch the source code](https://github.com/staltz/manyverse/blob/3e6f756d8e67e19aefd7eb158718c4168bd893d5/tools/patch-backend-ssb-server#L29) of ssb-db to replace mdmanifest usage, because it does a `fs.readFileSync` on non-JavaScript files expected to be in node_modules, but those files are deleted because we run `noderify` and delete all other files, in order to drastically decrease the amount of bytes shipped to the user's app download or update.

Another justification for this is that mdmanifest doesn't help that much, but creates "two sources of truth" which need to be synchronized. JavaScript is not the single authority concerning the definition of a muxrpc API, but also Markdown is. So if you (say) rename `createHistoryStream` to `createHistory`, you have to change that in both JS and api.md. Ideally, we should have one source of truth: either (1) JS is fully generated from markdown, or (2) markdown docs are generated from JS. (2) is actually quite common, it's doable through JSDocs for instance. There are several tools out there to generate markdown from JS, see for instance [markdox](https://github.com/cbou/markdox).

So this PR just removes mdmanifest as a dependency, and hardcodes the manifest in index.js.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,21 +4,39 @@ var path       = require('path')
 var osenv      = require('osenv')
 var mkdirp     = require('mkdirp')
 var rimraf     = require('rimraf')
-var mdm        = require('mdmanifest')
 var valid      = require('./lib/validators')
 var pkg        = require('./package.json')
 
 function isString(s) { return 'string' === typeof s }
 function isObject(o) { return 'object' === typeof o }
 function isFunction (f) { return 'function' === typeof f }
-// create SecretStack definition
-var fs = require('fs')
-var manifest = mdm.manifest(fs.readFileSync(path.join(__dirname, 'api.md'), 'utf-8'))
 
-manifest.seq = 'async'
-manifest.usage = 'sync'
-manifest.clock = 'async'
-manifest.version = 'sync'
+var manifest = {
+  get: 'async',
+  createFeedStream: 'source',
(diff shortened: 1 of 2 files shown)


arj03:
Sounds good to me.

I found two other unused references:

https://github.com/ssbc/ssb-plugins/blob/a51eadf39ddd27bdcf35aac9d5907ae1388039b7/index.js#L14

https://github.com/ssbc/ssb-invite/blob/2a0c98872bbda035fd9e2e6179da3ea19119d171/index.js#L8

index.js
```diff
@@ -4,16 +4,36 @@ var path       = require('path')
...
+  latest: 'source',
+  latestSequence: 'async',
+  whoami: 'sync',
+  del: 'async',
+  progress: 'sync',
+  status: 'sync',
+  getVectorClock: 'async',
+  version: 'sync',
+  help: 'sync'
+}
```
staltz: I should probably also change the lines below?

dominictarr:
:+1: go ahead please merge

PR #268, opened 2019-11-18
Add secret-stack name for simpler debugging

christianbundy:
Right now we aren't exporting a name, so it's hard to introspect which plugin this is without looking at the manifest.

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,6 +39,7 @@ var manifest = {
 }
 
 module.exports = {
+  name: 'db',
   manifest: manifest,
   permissions: {
     master: {allow: null, deny: null},


staltz:
Obvious merge

christianbundy:
Wait this might have broken everything. I think this moved our APIs, so `ssb.createHistoryStream()` is now `ssb.db.createHistoryStream()` and everything is broken. I'm going to revert this...

staltz:
Not obvious merge

dominictarr:
oh yeah, this is a historical quirk that we can't really change now. :crying_cat_face:

christianbundy:
@dominictarr Do you have thoughts about this plan from #274?

> I'm planning to add a backward-compat plugin as a sibling that exports everything with deprecation stack traces (except `createHistoryStream()`, of course).

I know that `createHistoryStream()` will need to exist as a top-level method, but I think we could move the others to a proper namespace as long as we add a deprecation notice early and make sure the removal happens in a major version bump.

dominictarr:
does this mean that `publish` now needs to be `db.publish` or you get a warning?

dominictarr:
I'm not against moving these to db... but I think adding a deprecation message for everything is unnecessary. it impacts everyone else who is working on an implementation. users start moving using db.X instead of X because of the annoying message, but that doesn't work with the other implementations. We shouldn't make changes like that without consulting the other implementors.

christianbundy:
I think that methods that are expected by other implementations should remain where they're at, but I don't think it's a bad idea to add deprecation notices for the other methods if we're going to move them to a `db` namespace. I know that `createHistoryStream()` is standard across all SSB implementations, but I'm not aware of the others.

staltz:
Deprecation warnings can also be throttled to appear every (e.g.) 20 usages of the method.

dominictarr:
@christianbundy maybe it's time collect a list of methods that are provided by the various implementations and in use by the clients (also tests because rewriting the tests is a pain)
lets not remove things without gauging the impact.

PR #270, opened 2019-11-20
fix README example var references

Connoropolous:
there were fragmented references to the `sbot` instance, named as `sbot`, `feed`, and `ssb` ... rename them all to `sbot`. 

Also add backticks to module names, and fix API formatting

See here for the updated README not as raw markdown: https://github.com/ssbc/ssb-db/blob/729cb08a91e4b7f08cb7bfa1b7c193f14f51d76d/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ sbot.whoami(function (err, data) {
 //  - feed.add appends a message to your key's chain.
 //  - the `type` attribute is required.
 
-feed.publish({ type: 'post', text: 'My First Post!' }, function (err, msg, hash) {
+sbot.publish({ type: 'post', text: 'My First Post!' }, function (err, msg, hash) {
   // the message as it appears in the database:
   console.log(msg)
 
@@ -57,7 +57,7 @@ feed.publish({ type: 'post', text: 'My First Post!' }, function (err, msg, hash)
 
 // stream all messages for all keypairs.
 pull(
-  ssb.createLogStream(),
+  sbot.createLogStream(),
   pull.collect(function (err, ary) {
     console.log(ary)
   })
@@ -65,7 +65,7 @@ pull(
 
 // stream all messages for a particular keypair.
 pull(
-  ssb.createHistoryStream({id: sbot.id}),
+  sbot.createHistoryStream({id: sbot.id}),
   pull.collect(function (err, ary) {
     console.log(ary)
   })
--- a/README.md
+++ b/README.md
@@ -24,22 +24,22 @@ that reads from the feed.
  */
 
 //create a secret-stack instance and add ssb-db, for persistence.
-var createSbot = require('secret-stack')({})
+var createSsbServerWithDb = require('secret-stack')({})
   .use(require('ssb-db'))
 
 
 // create the db instance.
 // Only one instance may be created at a time due to os locks on port and database files.
 
-var sbot = createSbot(require('ssb-config'))
+var ssbServerWithDb = createSsbServerWithDb(require('ssb-config'))
 
 //your public key, the default key of this instance.
 
-sbot.id
+ssbServerWithDb.id
 
 //or, called remotely
 
-sbot.whoami(function (err, data) {
+ssbServerWithDb.whoami(function (err, data) {
   console.log(data.id) //your id
 })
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks for this patch! I think most of this is right, but I don't think we should rename things back to "sbot" or "scuttlebot".

Some relevant context: this repo used to be called *secure-scuttlebutt* and the ssb-server used to be called *scuttlebot* (sometimes called *sbot* for short), but those names didn't really describe what the repos were. This repository is a database interface with lots of methods to query the database, and ssb-server is an secret-stack distribution that incorporates ssb-db with some other useful plugins (plus a CLI tool). I think we want to drop the name "sbot" in general.

Maybe instead of `ssbDb` in the docs we should just say `db`?

By the way, I'd love to get on a phone call with you and chat about all this stuff if you have any questions you'd like answered or things you'd like to discuss. I really appreciate all the hard work you've been doing to improve SSB docs lately.

Connoropolous:
> Maybe instead of ssbDb in the docs we should just say db?

Yep, I can make that change. It did seem a bit odd to me. I just wanted to make sure they all say the same thing, don't care what that thing is, as long as it's the best thing we can come up with!

README.md
```diff
@@ -24,30 +24,30 @@ that reads from the feed.
...
 
 //your public key, the default key of this instance.
 
-sbot.id
+ssbServerWithDb.id
 
 //or, called remotely
 
-sbot.whoami(function (err, data) {
+ssbServerWithDb.whoami(function (err, data) {
```
Connoropolous: @christianbundy do you know if `whoami` comes with `ssb-db` or something else?

christianbundy: Yep, it comes from ssb-db!

README.md
```diff
@@ -24,30 +24,30 @@ that reads from the feed.
  */
 
 //create a secret-stack instance and add ssb-db, for persistence.
-var createSbot = require('secret-stack')({})
+var createSsbServerWithDb = require('secret-stack')({})
```
christianbundy: Sorry to nit, but I think this might give the impression that it's creating an ssb-server instance.

Really, it's making an instance of secret-stack, which we're loading the ssb-db plugin into. For consistency with the [secret-stack docs](https://github.com/ssbc/secret-stack/#example) maybe this should be `createApp()`?

There's also some other weird context where most secret-stack plugins namespace their methods, but this plugin is naughty and makes top-level methods. In a consistent universe we'd do something like:

```javascript
const createApp = require('secret-stack')({})
  .use('ssb-db')

const app = createApp(require('ssb-config')

app.db.createHistoryStream()
//  ^ note the `db` here, which we don't actually do in the real world because we're inconsistent
```

Connoropolous:
@christianbundy if you like that last commit, then this pr is good to merge

README.md
Connoropolous: ah, true. No, these distinctions are important, I'm into it.

README.md
```diff
@@ -174,15 +174,16 @@ database does not have the private key:
...
 Without other plugins, this instance will not have replication
-or querying. Loading ssb-db directly is useful for testing,
+or querying. Loading `ssb-db` directly is useful for testing,
 but it's recommended to instead start from a plugin bundle like [ssb-server](https://github.com/ssbc/ssb-server)
 
-Because of legacy reasons, all the ssb-db methods are mounted on the top level object,
-so it's `sbot.get` instead of `sbot.db.get` as it would be with all the other `ssb-*` plugins.
+> Because of legacy reasons, all the `ssb-db` methods are mounted on the top level object,
+so it's `ssbServerWithDb.get` instead of `ssbServerWithDb.db.get` as it would be with all the other `ssb-*` plugins.
+> In the API docs below, we'll just call it `db`
```
Connoropolous: @christianbundy I added an extra note about that here, but I'll change the `ssbServerWithDb` reference

README.md
christianbundy: If you're down with this change I'm happy to make the edit so you don't have to add another commit, but I wanted to make sure I discussed with you because you have a much better perspective on the documentation. Working in an ecosystem for too long warps your brain and it makes everything feel "normal", which is often really bad for writing good documentation.

README.md
christianbundy: Oh nice! Totally slipped my brain.

christianbundy:
This looks really great, thanks for improving these docs so much! So much attention to detail.

PR #273, opened 2019-11-22
Remove non-existed methods from muxrpc manifest

### LINKED ISSUE #272: clock() exposed over muxrpc but isn't exported?
I'm building a tiny SSB CLI and I noticed that this command is throwing an error locally.

christianbundy:
Resolves #272 

We weren't actually exporting this so when experimenting with [ssb-cli](https://github.com/fraction/ssb-cli) there were lots of broken methods showing up.

### DIFF
--- a/help.js
+++ b/help.js
@@ -18,10 +18,6 @@ function rangeArgs (type) {
   }
 }
 
-function isTimestamp (arg) {
-  return !isNaN(+arg) && +arg >= 0
-}
-
 var keysAndValues = {
    keys: {
       type: 'boolean',
@@ -37,12 +33,13 @@ var keysAndValues = {
 
 var MessageId = {
   type: 'string',
-  test: /^%[a-zA-Z0-9\+\/]+={0,2}\.\w+$/
+  test: /^%[a-zA-Z0-9+/]+={0,2}\.\w+$/
 }
 
 var Private = {
   type: 'boolean',
-  description: 'decrypt private messages, defaults to false'
+  description: 'decrypt private messages',
+  default: false
 }
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


christianbundy:
I've also fixed up `help.js`, which was giving incorrect usage info about whether `--seq` was required.

help.js
```diff
@@ -37,12 +33,13 @@ var keysAndValues = {
 
 var MessageId = {
   type: 'string',
-  test: /^%[a-zA-Z0-9\+\/]+={0,2}\.\w+$/
+  test: /^%[a-zA-Z0-9+/]+={0,2}\.\w+$/
```
christianbundy: I removed these escapes because they aren't actually doing anything in this context.

help.js
```diff
@@ -107,7 +105,7 @@ module.exports = {
         },
         seq: {
           type: 'SequenceNumber',
-          optional: false,
+          optional: true,
```
christianbundy: Fixing this unbreaks [`ssb createHistoryStream`](https://github.com/fraction/ssb-cli#usage). :tada:

index.js
```diff
@@ -22,7 +22,6 @@ var manifest = {
   links: 'source',
   add: 'async',
   publish: 'async',
-  getAddress: 'sync',
```
mixmix, line 25: interestingly, this was not in `v19.1.1`

I've been stepping back through time looking for getAddress references: 
```
$ git checkout HEAD~1 && ack getAddress
```

mixmix: adding this back in does not seem to help with the `ssb-server` problem I'm currently having which complains this method doesn't exist

PR #274, opened 2019-11-22
Revert "Add secret-stack name for simpler debugging"

christianbundy:
Reverts #268 because it breaks ssb-db via muxrpc. I'm planning to add a backward-compat plugin as a sibling that exports everything with deprecation stack traces (except `createHistoryStream()`, of course).


christianbundy:
Merging as "obvious".

PR #275, opened 2019-11-25
Export createSequenceStream() for client-side Secret Stack plugins

christianbundy:
This should probably start out as a discussion, but I'd love to support client-side Secret-Stack plugins by exporting the FlumeDB log over MuxRPC. I'd love to hear feedback on this idea, and if it's something we all like then I'd like to move forward with this PR.

SSB thread with more context: [`%LiQ89SmUZJU7gRZfTLJzLUrsCO2jMyXeWcMufSqq50Q=.sha256`](http://oasis-demo.fraction.io/thread/%25p2pQbjaNYrKYNhQOXTjhjdQLoDZPcQpEMtOpwJjn1%2FA%3D.sha256)

This convention would probably have consequences for other implementations, so I've pinged lots of folks I think would care about this.

Basic (JS) usage might look like:

```javascript
client((err, ssb) => {
  const local = magic(ssb)
    .use(require('ssb-backlinks'))
    .use(require('ssb-about'))()

  // Basic 'about' message.
  const content = {
    type: 'about',
    about: '@xyz',
    name: 'Xander'
  }

  ssb.publish(content, () => {
    // Sometimes there's a race condition here.
    // Maybe we should change `publish()` so it doesn't callback immediately.
    local.about.socialValue({ key: 'name', dest: '@xyz' }, (err, foundName) => {
      console.log(foundName) // => "Xander"
    })
  })
})
```


christianbundy:
On second thought, it might be easier to just:

- Add `{ seq }` support to `createLogStream()`
- Expose the `since` observable over MuxRPC as a source of `seq`s

dominictarr:
you can get raw access to the log using `createLogStream({raw:true})` https://github.com/ssbc/ssb-db/blob/9feb2d21c9e9147043531f29d66003355e122a97/extras.js#L17 , in which case gt,lt now refer to the sequence. a source stream for `since` would be a good idea...

maybe should make a createFlumeLogStream or something though, instead of an option that changes how the whole thing works, that's not nice if you are in a strictly typed language.

I tried client side views before, such as the remote version of flumeview-reduce https://github.com/flumedb/flumeview-reduce/blob/master/store/remote.js I thought it was gonna be faster to load but it was slower. Really I think the best way to do this would be instead of having a client side view running, would be to send the code for the view to the server, and run it inside secure ecmascript or wasm.

That said, views in separate processes (such as the client) have always been a part of my mental model for flume, so I don't want to stand in the way of this.

christianbundy:
Thanks @dominictarr! I used that to reduce the changes in this PR, so now the only API addition is `sinceStream`. Is pull-pushable the right tool for this job? Or is there a better way to create a client-side `since` observable that matches flumedb's `since` observable?

I wouldn't be surprised if there's some on-zero performance cost, but I think it's kind of cool that all we need is a `since` stream and we've got client-side secret-stack plugins that only depend on ssb-db.

index.js
```diff
@@ -6,6 +6,8 @@ var mkdirp     = require('mkdirp')
 var rimraf     = require('rimraf')
 var valid      = require('./lib/validators')
 var pkg        = require('./package.json')
+const pullPushable = require('pull-pushable')
+const p = pullPushable()
```
staltz: Can we name this better than one letter? Perhaps "sincePushable"

index.js
```diff
@@ -93,6 +96,12 @@ module.exports = {
       }
     }
     var self
+
+
+    ssb.since((value) => {
+      p.push(value)
```
staltz: pull-pushable uses a buffer until there are consumers of the stream. So if it turns out that `sinceStream` is not used anywhere, pull-pushable will still buffer all the log in an array in memory! :warning:

I think we need more careful consideration on this implementation. Perhaps it should be a lazy source, that only begins reading the log once the source is pulled for the first time.

index.js
staltz: Also, could move the declaration of this. Out of the "import" block and into the plugin's `init`.

christianbundy: Done, thanks!

index.js
christianbundy: No good! I just switched over to [pull-notify](https://github.com/pull-stream/pull-notify/), which seems to resolve this by **creating one stream per consumer** and **only pushing to active streams**. Does that seem more reasonable to you?

christianbundy:
Thinking about this again, maybe I should just expose `createRawLogStream()` over MuxRPC and then just call it with `{ live: true, values: false }`? It'd be a smaller change with no new API.

christianbundy:
Scratch that, I can't find a way to tell `createRawLogStream()` to give me a source that returns **only the latest seq** plus all future sequence numbers. I think the source stream is the way to go.

christianbundy:
Dumping more info here, I built ssb-backlinks from scratch with my full log and it looks like it's 5.5x faster to do a server-side view:

- **Server-side:** 4 minutes 21 seconds
- **Client-side:** 22 minutes 41 seconds

My guess is that it's the overhead from the muxrpc stream, but maybe it's something else? Might be worth trying with noauth over a unix socket.

dominictarr:
@christianbundy to build a view, we care about only the latest, but not about things before that. So if we start the stream, and there is already a large log, we just want a single item. That's easy. But then if a big amount of items get added, we don't really need to have every new since value sent, just want to know that it increases to... for example, say we build an app using this and then do an initial sync, if we simply stream each new since value we'll send every single one. But we are adding items really fast, so if we only sent the biggest one every 200 ms, then we'd send a lot less and it would still behave correctly.

It's interesting that noauth is nearly as fast! they both use muxrpc, but the main difference is pull-box-stream

dominictarr:
@christianbundy you should also look at overall memory use, really we need to graph that or something. In a perfect world, we could some use available working memory to set the back pressure, if you start to run out of memory and go into swap space it will slow down a lot! Also, in a single threaded enviroment (well, in node.js at least) hogging the cpu too much can interfere with receiving net sockets, so that should create back pressure too...

staltz:
Exciting that unix~noauth is nearly the same as before with the plugin on the server. I think the specifics of the client-side performance of each plugin can be debated, but overall the possibility of doing it client-side is promising. We also could leverage JavaScript "threads" such as WebWorkers or Node.js [tiny-worker](https://www.npmjs.com/package/tiny-worker) or Electron renderers. By the way I made [pull-worker](https://github.com/staltz/pull-worker) once.

index.js
```diff
@@ -113,13 +124,7 @@ module.exports = {
         return pkg.version
       },
 
-      //temporary!
-      _flumeUse                :
-        function (name, flumeview) {
-          ssb.use(name, flumeview)
-          return ssb[name]
-        },
-
```
staltz: Removing `_flumeUse` is a big deal, right? A lot of plugins use it, so this is a breaking change? Do tests fail? And I never really understood why was it marked "temporary" and why does it start with an underscore.

christianbundy: Yeah this should go back in -- this PR is super messy because it was meant to open the discussion with a hacky prototype that I wrote. I think that *really*, this PR should be:

- A new method added to `index.js` that creates a stream of sequence numbers
- A few tests added to ensure that the sequence stream is working as intended

Currently this PR has a bunch of experimental code running client-side flumeviews and stuff, but I don't think we actually want that in ssb-db.

christianbundy:
Just reverted all of the experimental stuff, now it's just `createSequenceStream()` and a test file.

> But we are adding items really fast, so if we only sent the biggest one every 200 ms, then we'd send a lot less and it would still behave correctly.

Yeah, and I experimented with pull-debounce to try to avoid sending values so often but ran into a problem: it makes ssb-db artificially slow when *not* running over ssb-db, so after `ssb.publish()` you have to wait 200 ms (!) until your observable notices the change. I'm tempted to suggest that muxrpc could resolve this by supporting a special "debounce-able source" or something, since the debounce timing is based on the round-trip latency, but I don't think we need to optimize it yet.

> you should also look at overall memory use, really we need to graph that or something.

:+1:

When I have time to explore some of this mad science a bit more I'll probably need to do some research and find the bottlenecks. I'm curious whether we could optimize by reusing an existing stream for views that need to index the same messages, but again -- I don't think that needs to be optimized yet. Baby steps!

dominictarr:
@staltz it was going to be temporary, which is why it started with _, but then it never got changed.

christianbundy:
Is there anything else we should discuss in this PR?

test/sequence-stream.js
```diff
@@ -0,0 +1,50 @@
...
+
+  ssb.publish({type: 'test'}, (err, publishedMessage) => {
+    t.error(err, 'publish() success')
+
+    t.equal(since.value, 0, 'since is incremented')
+    t.end()
+  })
+})
+
+test('resume sequence strem later', (t) => {
```
staltz: strem => stream

staltz:
Could we get a readme documentation for this new API?

test/sequence-stream.js
christianbundy: Thanks!

christianbundy:
Yep, just pushed some commits. Please let me know if there's anything else.

staltz:
To me, this looks good to merge but I think it would be wise to have the approval from one more person, at least.

christianbundy:
@staltz Is there anyone in particular you'd like to review this? It's a pain point to keep so many pull requests open but I'm not sure how to resolve it.

staltz:
@arj03 ?

arj03:
Ohh why havn't I seen this thread before! This is quite interesting. Thanks for the ping @staltz. I think the api looks good, so I approve this PR. I need to check out pull-worker, didn't know that existed.

PR #276, opened 2019-12-03
Change Travis tests to replace Node 6 with Node 12

christianbundy:
Node 8 is almost at EOL, so I think it's probably fine to finally drop Node 6 support in favor of the active LTS.


arj03:
👍

PR #280, opened 2020-01-28
Add boxer

mixmix:
In working on private groups I'm keen to register new boxers (as well as unboxers).

This PR adds:
- adds `db.addBoxer` which follows a similar pattern to `addUnboxer`
- moves private-box (box1) code to use the new API (in preparation to extract it!)

### DIFF
--- a/create.js
+++ b/create.js
@@ -2,11 +2,12 @@
 
 var join = require('path').join
 var EventEmitter = require('events')
-
 var pull = require('pull-stream')
 var ref = require('ssb-ref')
 var ssbKeys = require('ssb-keys')
 
+var createDB = require('./db')
+var extras = require('./extras')
 var u = require('./util')
 
 function isString (s) {
@@ -17,13 +18,13 @@ function errorCB (err) {
   if (err) throw err
 }
 
-module.exports = function (path, opts, keys) {
+module.exports = function create (path, opts, keys) {
   //_ was legacy db. removed that, but for backwards compatibilty reasons do not change interface
   if(!path) throw new Error('path must be provided')
 
   keys = keys || ssbKeys.generate()
 
-  var db = require('./db')(join(opts.path || path, 'flume'), keys, opts)
+  var db = createDB(join(opts.path || path, 'flume'), keys, opts)
 
   // UGLY HACK, but...
   // fairly sure that something up the stack expects ssb to be an event emitter.
@@ -140,7 +141,7 @@ module.exports = function (path, opts, keys) {
 
   // pull in the features that are needed to pass the tests
   // and that sbot, etc uses but are slow.
-  require('./extras')(db, opts, keys)
+  extras(db, opts, keys)
 
   // writeStream - used in (legacy) replication.
   db.createWriteStream = function (cb) {
@@ -185,7 +186,3 @@ module.exports = function (path, opts, keys) {
 
   return db
 }
-
-
-
-
--- a/index.js
+++ b/index.js
@@ -106,7 +106,11 @@ module.exports = {
       },
 
       status                   : function () {
-        return {progress: self.progress(), db: ssb.status, sync: since() }
+        return {
+          progress: self.progress(),
+          db: ssb.status,
+          sync: since()
+        }
       },
 
       version                  : function () {
@@ -114,17 +118,16 @@ module.exports = {
       },
 
       //temporary!
-      _flumeUse                :
-        function (name, flumeview) {
-          ssb.use(name, flumeview)
-          return ssb[name]
-        },
+      _flumeUse                : function (name, flumeview) {
+        ssb.use(name, flumeview)
+        return ssb[name]
+      },
 
       close                    : close,
-      del: valid.async(ssb.del, 'msgLink|feedId'),
+      del                      : valid.async(ssb.del, 'msgLink|feedId'),
       publish                  : valid.async(feed.add, 'string|msgContent'),
       add                      : valid.async(ssb.add, 'msg'),
-      queue                      : valid.async(ssb.queue, 'msg'),
+      queue                    : valid.async(ssb.queue, 'msg'),
       get                      : valid.async(ssb.get, 'msgLink|number|object'),
 
       post                     : ssb.post,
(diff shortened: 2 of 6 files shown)


staltz:
I like what I see but I haven't reviewed it yet, and being about privacy and touching ssb-db, this needs special attention. Will (probably) come back to this and review properly.

index.js
```diff
@@ -106,25 +106,28 @@ module.exports = {
...
+        return ssb[name]
+      },
 
       close                    : close,
-      del: valid.async(ssb.del, 'msgLink|feedId'),
+      del                      : valid.async(ssb.del, 'msgLink|feedId'),
       publish                  : valid.async(feed.add, 'string|msgContent'),
       add                      : valid.async(ssb.add, 'msg'),
-      queue                      : valid.async(ssb.queue, 'msg'),
+      queue                    : valid.async(ssb.queue, 'msg'),
```
mixmix: sorry, these misaligned dots have been eating my brain for 1-2 years.
I couldn't leave them alone

create.js
```diff
@@ -2,11 +2,12 @@
 
 var join = require('path').join
 var EventEmitter = require('events')
-
 var pull = require('pull-stream')
 var ref = require('ssb-ref')
 var ssbKeys = require('ssb-keys')
 
+var createDB = require('./db')
+var extras = require('./extras')
```
mixmix: I pulled these to the top of the file because I like to see what's been pulled into a file
I was passing through here trying to understand where the addUnboxer connected in an how.
I hope this help later people reading to read more easily

test/end-to-end.js
```diff
@@ -99,11 +95,11 @@ module.exports = function (opts) {
...
           // bob can also decrypt
           var content = ssbKeys.unbox(ctxt, alice.private)
-          t.deepEqual(content, {
-            type: 'secret2',
-            secret: "it's a secret!",
-            recps: [alice.id]
-          }, 'alice can decrypt')
+          t.deepEqual(
+            content,
+            { type: 'secret2', secret: "it's a secret!", recps: [alice.id] },
```
mixmix: :fire: this test does something very strange (I think).
if you compare what was published with `feed.add` and what comes out the the private box ... they're different!

The RECPS field has been array'd !
This is fine, so long as it doesn't mess with signature / hashes of content?
Maybe I don't understand the order of that relative to encryption though

mixmix: _I didn't change the test, just the layout to make it easy to compare. The code I wrote still passes this strange test_

create.js
christianbundy: Very much appreciated, thank you.

minimal.js
```diff
@@ -208,34 +221,48 @@ module.exports = function (dirname, keys, opts) {
...
-      } else {
-        throw new Error(
-          'private message recipients must be valid feeds, '+
-          'was:' + JSON.stringify(recps)
-        )
-      }
+  /* TODO extract to ssb-private */
+  var box1 = {
+    box: function (content, recps) {
+      if (!recps.every(isFeed)) return null
```
christianbundy: IIRC Patchbay sets `recps` as `{ link, name }`, is that something we want to support here?

minimal.js
```diff
@@ -208,34 +221,48 @@ module.exports = function (dirname, keys, opts) {
...
+  db.addUnboxer(box1.unbox)
+  // ////////////////////////////////
+
+  // db.box = function (value, state) {
+  db.box = function (value) {
+    var recps = value.content.recps
+    if (!recps) return value.content
+
+    if (typeof recps === 'string') recps = value.content.recps = [recps]
+    if (!Array.isArray(recps)) throw new Error('private message field "recps" expects an Array of recipients')
```
christianbundy: Maybe redundant, but my intuition is that this change would throw errors when Patchbay makes private messages. If that's true, it would be important to ensure that this is published as a breaking change. FWIW, I'm a fan of the simplicity.

minimal.js
```diff
@@ -208,34 +221,48 @@ module.exports = function (dirname, keys, opts) {
...
+  db.addBoxer(box1.box)
+  db.addUnboxer(box1.unbox)
+  // ////////////////////////////////
+
+  // db.box = function (value, state) {
+  db.box = function (value) {
+    var recps = value.content.recps
+    if (!recps) return value.content
+
+    if (typeof recps === 'string') recps = value.content.recps = [recps]
```
christianbundy: I'm comfortable with this, but would also be comfortable not supporting this use-case and just insisting that `recps` is an array.

minimal.js
mixmix: I don't want to support everyone and their dog. I just want it to be an Array of `feedIds`. Keep in mind this is just the boxer, it doesn't change anything already published.

Also, I'm _fairly_ sure my code doesn't change any behaviour here...

minimal.js
mixmix: yeah me too. I was trying to make this a PR added an API, but didn't change any behaviour...

minimal.js
mixmix: Hey patchbay stuff was written before the magical `recps` auto-encryption (I think). Here's the code in patchcore which is doing the encryption for patchbay @christianbundy : https://github.com/ssbc/patchcore/blob/master/sbot.js#L163-L168

TL;DR - this isn't a breaking change

minimal.js
mixmix: I don't see how this make a breaking change. See above comment

minimal.js
christianbundy: Sweet! Thanks for learning me some history, I don't remember `recps` auto-encryption even being a new thing, but that's great info to have.

minimal.js
christianbundy: Yup, my bad there.

create.js
```diff
@@ -17,13 +18,13 @@ function errorCB (err) {
...
 
-module.exports = function (path, opts, keys) {
+module.exports = function create (path, opts, keys) {
   //_ was legacy db. removed that, but for backwards compatibilty reasons do not change interface
   if(!path) throw new Error('path must be provided')
 
   keys = keys || ssbKeys.generate()
 
-  var db = require('./db')(join(opts.path || path, 'flume'), keys, opts)
+  var db = createDB(join(opts.path || path, 'flume'), keys, opts)
```
arj03: This is much nicer to read ❤️

mixmix:
@christianbundy @arj03 is this good enough to merge in your opinion.

Should I run other tests like symlink in ssb-server and see if breaks all those integrations?

christianbundy:
The SSB-Server tests are consistenty broken for me every time I try them (including tonight). If you have a way of easily testing, that'd be cool, but I'd be comfortable just optimistically merging this.

arj03:
I think if the tests passes for this repo it should be fine to merge.

mixmix:
Published as `ssb-db@19.4.0`

mixmix:
Thanks for help / input on this @arj03 + @christianbundy

PR #286, opened 2020-03-10
Upgrade FlumeDB

christianbundy:
Problem: The old version of FlumeDB can't auto-heal broken indexes.

Solution: Upgrade FlumeDB to ^2.1.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "async-write": "^2.1.0",
-    "flumedb": "^1.1.0",
+    "flumedb": "^2.1.1",
     "flumelog-offset": "^3.4.2",
     "flumeview-level": "^3.0.13",
     "flumeview-reduce": "^1.3.9",


christianbundy:
This is a breaking change because it requires that views implement the `.close()` method, if I understand correctly. I think we should probably release as a major version bump so that people don't accidentally upgrade and get errors.

Powersource:
I don't know how I would go about reviewing this, got any pointers?

Not knowing this module very well, I would say just merge this, since tests are passing? Should we maybe do a proof of concept upgrade of a view to know that that works too?

christianbundy:
Our views, Flumeview-Level and Flumeview-Reduce, both implement the `close()` method so I think this module is good to go. I'd like to merge, bump the major version, and then leave it up to modules that depend on SSB-DB to upgrade when they can. My understanding is that FlumeDB should now throw an error on startup if there's no `close()` method, so worst case scenario someone tries to update and when they `npm start` or `npm test` they get an error.

Powersource:
Questions, because I'm curious,

> Our views, Flumeview-Level and Flumeview-Reduce

By 'our', do you mean ssb(c)? I thought `view == index` in this context but maybe I've misunderstood? If they're `!=`, where do the other ssb indexes come in?

christianbundy:
Sorry, by "our" I mean SSB-DB's views. I think we can safely upgrade this module, it's just that since this module *also* exports `_flumeUse()` then we need to do a major version bump whenever FlumeDB does.

PR #288, opened 2020-05-20
Fix `test/close.js` on async-box with `{ temp }`

christianbundy:
Problem: The `test/close.js` test was failing because the database
seems to have been empty, and since `createLogStream()` doesn't emit
items on an empty database then `pull.collect()` is never called. The
database is empty because the first test creates a database where `temp`
is implicitly true, which uses a directory like `/tmp/1590008707854`.
When the second test sets `temp` to false, it uses a named directory
like `/tmp/test-ssb-feed`, which is **empty**.

Solution: Pass `{ temp: false }` consistently to both tests.

### DIFF
--- a/test/close.js
+++ b/test/close.js
@@ -26,7 +26,7 @@ tape('reopen', function (t) {
   t.plan(1)
 
   // HACK: See readme section on 'known bugs'.
-  setImmediate(() => {
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


christianbundy:
Oh interesting, this only works the *second* time you run it, because the data is cached to `/tmp`. Hmm.

PR #292, opened 2020-05-26
Add rebuild method

christianbundy:
Problem: There's no way to rebuild the database from SSB-DB, which
is necessary when you get a new unboxer key or want to resolve some
weird database bug under the hood.

Solution: Add `rebuild()` method to database and add tests to ensure
that it has the expected behavior. Currently it does **not**, and this
suggests that FlumeDB's rebuild code isn't actually doing any rebuilding
under the hood.

---

Minor version bump. Depends on:

- https://github.com/flumedb/flumeview-level/pull/25
  - https://github.com/flumedb/test-flumeview-index/pull/3
- https://github.com/flumedb/flumedb/pull/46


index.js
```diff
@@ -15,27 +15,28 @@ function isObject(o) { return 'object' === typeof o }
...
+  messagesByType: 'source',
   progress: 'sync',
+  publish: 'async',
+  rebuild: 'async',
   status: 'sync',
-  getVectorClock: 'async',
   version: 'sync',
-  help: 'sync',
+  whoami: 'sync',
 }
```
christianbundy, line 40: Sorted and added 'rebuild'.

test/create-ssb.js
```diff
@@ -1,12 +1,20 @@
...
-    require('mkdirp').sync(dir)
+    rimraf.sync(dir);
+    mkdirp.sync(dir);
   }
-  var keys = opts.keys = opts.keys || ssbKeys.generate()
-  return require('../create')(dir, opts, keys)
-}
+
+  const keys = (opts.keys = opts.keys || ssbKeys.generate());
+  return require("../create")(dir, opts, keys);
```
mixmix, line 19: missed one!

PR #293, opened 2020-06-03
put an LRU cache over unbox

mixmix:
```
/* NOTE
 * currently when flumedb passes messages to each view (index) it runs
 * an unbox on encrypted messaged FOR EVERY VIEW
 *
 * This is a temporary easy solution to reduce some wasted CPU cycles
 * without having to change deep things about flumedb
 */
```

This PR uses an LRU style cache to _bound_ the amount of the memory this unbox cache will use.

This needs to take into account index rebuilds in the context where you've just been given a private group key. This is why there's (untested) code which closes over the `flume.rebuild` function

### DIFF
--- a/autobox.js
+++ b/autobox.js
@@ -1,3 +1,4 @@
+const HLRU = require('hashlru')
 const { metaBackup } = require('./util')
 
 function isFunction (f) { return typeof f === 'function' }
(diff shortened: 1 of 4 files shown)


mixmix:
Note this is a very late night gesture at a solution ... I've seen parts of it working but expect it to be part-baked

mixmix:
Sigh, travis says _YES!_, ...but  it's only testing on Linux... and this is another one of those repos with a unix for-loop which kills windows?  (and it's got those utils / test files ... not sure how intense they are)

mixmix:
PR for tests on all platforms: https://github.com/ssbc/ssb-db/pull/297

mixmix:
tested with base indexes in `ssb-db`, result: 
- cuts indexing from 4.5min to 3.5min

I'm gonna merge and increase the cache size because I notice the views get out of sync, which then reduces the effectiveness in rebuild

PR #296, opened 2020-06-16
upgrade mkdirp

mixmix:
there was some security warning about minimist which mkdirp depends on.
noticed there was a major version with a simialr API (for our purposes, so bumped)


christianbundy:
Passes tests for me, thanks for the upgrade!

PR #298, opened 2020-06-16
Add full tests for index.js, not create.js

christianbundy:
Problem: Running tests that import from `create.js` means that we're not
actually testing the module that we're exporting to applications that
depend on SSB-DB. This means that our test have different behavior in
development versus production, and makes it easier for subtle bugs to
split in.

Solution: Test against `index.js`, not `create.js`, and fix all of the
problems that occur when you test against the full module instead of the
previous subset.

---

This takes us from 265 assertions and 72% code coverage to 273 assertions and 79% code coverage. :tada: 

Unfortunately, it didn't help me find the bug that I'm trying to catch, but that's okay.


christianbundy:
Self-merging because this only adds tests and fixes new bugs (:tada:), happy to revert if anyone would prefer an in-depth review.

PR #301, opened 2020-06-18
Add boxer init method

christianbundy:
Problem: Boxer init and unboxer init happen at different times, and
the constraint that they happen at the same time can lead to a circular
dependency where indexing becomes locked.

Solution: Add boxer init method.


test/box-unbox.js
```diff
@@ -324,12 +338,13 @@ function run () {
     const ciphertext = Buffer.from(JSON.stringify(content)).toString('base64') + '.box.hah'
 
     feed.publish(ciphertext, (_, msg) => {
-      t.true(initDone, 'unboxer completed initialisation before publish')
+      t.true(boxerInitDone, 'boxer completed initialisation before publish')
+      t.false(unboxerInitDone, 'unboxer did not completed initialisation before publish')
```
christianbundy, line 342: EXTRA ATTENTION HERE: I changed a test. I don't think I made a mistake, but changing tests is always kinda scary. :eyes: 

Previously we made sure that the unboxer init was done after publishing, but now that they're decoupled we don't want to enforce this constraint. Instead, we want to ensure that the **boxer** init is done, and (since I've made the unboxer init take twice as long) we want to ensure that the **unboxer** init is not done.

test/util/create-ssb.js
```diff
@@ -11,10 +11,14 @@ const caps = { shs: crypto.randomBytes(32).toString('base64') }
 
 const randomName = () => crypto.randomBytes(16).toString("hex");
 
-module.exports = function createSSB (name = randomName(), opts = {}) {
+module.exports = function createSSB (name = randomName(), opts = {}, plugins) {
```
christianbundy, line 14: Now `createSSB()` can add plugins! I'm using this to test SSB-Private2.

README.md
```diff
@@ -593,7 +593,11 @@ db.since(fn(seq)) => Obv
 An [observable](https://github.com/dominictarr/obv) of the current log sequence. This is always a positive 
 integer that usually increases, except in the exceptional circumstance that the log is deleted or corrupted.
 
-### db.addBoxer(boxer)
+### db.addBoxer: sync
+
+```js
+db.addUnboxer({ value: boxer, init: initBoxer })
```
mixmix, line 599: :fire: `init` is undocumented

README.md
```diff
@@ -605,9 +609,11 @@ automatically box (encrypt) the message `content` if the appropriate
 - break (because it should know how to handle `recps`, but can't), and so throw an `Error`
 
 ## db.addUnboxer: sync
+
 ```js
 db.addUnboxer({ key: unboxKey, value: unboxValue, init: initBoxer })
 ```
+
```
mixmix, line 616: this should probably change to `initUnboxer` for clarity (now that we have another `initBoxer`

minimal.js
```diff
@@ -144,7 +150,25 @@ module.exports = function (dirname, keys, opts) {
...
+    switch (typeof boxer) {
+      case 'function':
+        boxers.push(boxer)
+        break
+
+      case 'object':
+        if (typeof boxer.value !== 'function') throw new Error('invalid boxer')
+
+        if (boxer.init) {
+          setup.boxers.add(boxer.init)
```
mixmix, line 162: I think I made it so this could be a one-line `setup.boxers.add(boxer.init).runAll()` but not likely worth the change

minimal.js
```diff
@@ -144,7 +150,25 @@ module.exports = function (dirname, keys, opts) {
   }
 
   db.addBoxer = function addBoxer (boxer) {
-    boxers.push(boxer)
+    switch (typeof boxer) {
```
mixmix, line 153: `<style>` with function which have multiple signatures I think I prefer one main path like : 

```js
function addBoxer (boxer) {
  if (typeof boxer === 'function') return addBoxer({ value: boxer })
  // this makes it clear we're putting someone back on the *main* path, and reduces the nesting in the fn

  // ... the main path
}
```

test/box-unbox.js
mixmix: oh dear ... see comment below

test/private2.js
```diff
@@ -0,0 +1,30 @@
...
+  const ssb2 = createSSB(name, { temp: false }, [require('ssb-private2')])
+
+  const result = await promisify(ssb2.get)(key)
+  t.ok(result)
+
+  await promisify(ssb2.publish)({ type: 'test' })
+  t.pass('published again')
+
+  await promisify(ssb2.close)()
+  t.pass('closed again')
```
mixmix, line 29: I don't understand what this is testing exactly

test/util/create-ssb.js
mixmix: I think we've accidentally recreated `scuttle-testbot` ...?

mixmix: oh, except that thing has `ssb-db` as a dep

minimal.js
mixmix: I pushed this change because I have itchy fingers

christianbundy: Sweet, this method is just a copy and paste of the `addUnboxer()` method. Since `addUnboxer()` used the `switch...case` pattern I used the same here, but I think the important bit is that they're consistent. 🤷

test/private2.js
christianbundy: I don't completely understand why, but this was able to give us a minimal reproduction of the problem that we were bumping into with `whakapapa-ora/desktop/rebuild.temp.js`.

test/util/create-ssb.js
christianbundy: Oh yeah, it's super similar. I think the SSB-DB dep would make it hard to test in this module, since we want to load our own SSB-DB, but that's good to know for the future.

minimal.js
mixmix: yeah I saw that when I started editing. I tried changing `addUnboxer` to drop the switch but it's not possible there. The problem is that addUnboxer accepts 2 radically different modes - a onestep unbox or a two-step unbox. sigh

test/private2.js
mixmix: oh with the locked up state?

mixmix: I've left a note in that file so people looking at it have some context

README.md
mixmix: DONE

mixmix:
NOTE - I spent half a day working to get windows tests passing. I'm fairly confused about what's happening.

TL;DR
- `ssb.close` is calling back with an error about the server already being closed, but intermittent
- have tests node 10, 12 on and actual windows machine and it's fine ...

**MERGING**

PR #302, opened 2020-06-19
Boxer init ...tests

mixmix:
oh jeez this started as a little tidying / refactoring, and kinda expanded to involve adding tests to cover some untested things, and restructuring the tests to be "safer"

I am aware that reviewing this is probably annoying. For the most part I just moved tests around, but the changes will mean the diff is brutal I think.

Note:
- formally there were 84 tests in test/box-unbox.js, now there are 111
- the tests for that file used to run in ~2s, then now take ~4s
    - I think that's due to opening and closing multiple ssb instances (safer but slower)
    - I wonder if we should revert some of the changes in `util/create-ssb`. Same argument though right, slow, but better test?


test/box-unbox.js
```diff
@@ -233,189 +245,434 @@ function run () {
     })
   })
 
-  tape('addBoxer', function (t) {
-    const boxer = (content) => {
-      if (!content.recps.every(r => r === '!test')) return
+  tape.onFinish(ssb.close)
```
mixmix, line 248: I don't know any other way to close the ssb instance in this scope

mixmix:
wow this is confusing @github 
it's merged ... but I could merge it again huh?

mixmix:
ohhp, and when I publish a message you hide the merge button ... even though you didn't after I refreshed the page ... NICE

PR #303, opened 2020-06-22
refactor db.get

mixmix:
I was debugging something and was seeing an error in the neighbourhood of this method.
I got to logging and noticed some edge cases I could close

### DIFF
--- a/create.js
+++ b/create.js
@@ -46,32 +46,17 @@ module.exports = function create (path, opts, keys) {
   //   - db.getLatest
 
   db.get = function (key, cb) {
-    let isPrivate = false
-    let unbox
+    let doUnbox = false
+    let readKey
     let meta = false
     if (typeof key === 'object') {
-      isPrivate = key.private === true
-      unbox = key.unbox
+      doUnbox = key.private === true
+      readKey = key.unbox
       meta = key.meta
       key = key.id
     }
 
-    if (ref.isMsg(key)) {
-      return db.keys.get(key, function (err, data) {
-        if (err) return cb(err)
-
-        if (!isPrivate) {
-          if (meta) cb(null, { key, value: u.originalValue(data.value), timestamp: data.timestamp })
-          else cb(null, u.originalValue(data.value))
-        }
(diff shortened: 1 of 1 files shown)


create.js
```diff
@@ -46,42 +46,48 @@ module.exports = function create (path, opts, keys) {
...
-    if (ref.isMsg(key)) {
-      return db.keys.get(key, function (err, data) {
-        if (err) return cb(err)
-
-        if (!isPrivate) {
-          if (meta) cb(null, { key, value: u.originalValue(data.value), timestamp: data.timestamp })
-          else cb(null, u.originalValue(data.value))
-        }
-        else {
-          const result = db._unbox(data, unbox)
```
mixmix, line 68: this call to `_unbox` is un-needed in the case that the message was already auto-unboxed (by the chainmap that `db.keys` has in place).
we only need to call `_db.unbox` when we've been given the `readKey` for a message we couldn't previously decrypt.

Probably doesn't save that many cycles, but seemed silly to run it through there

create.js
```diff
@@ -46,42 +46,48 @@ module.exports = function create (path, opts, keys) {
...
-
-          if (meta) cb(null, { key, value: result.value, timestamp: result.timestamp })
-          else cb(null, result.value)
-        }
-      })
-    } else if (ref.isMsgLink(key)) {
```
mixmix, line 74: I moved this earlier for clarity (it's an early now an return)

create.js
```diff
@@ -46,42 +46,48 @@ module.exports = function create (path, opts, keys) {
...
+          else {
+            const unboxed = db._unbox(data, readKey)
+            if (unboxed) value = unboxed.value
+          }
+        }
+
+        if (!value) value = data.value // if db._unbox fails value will be undefined
+
+        if (meta) cb(null, { key, value, timestamp: data.timestamp })
+        else cb(null, value)
```
mixmix, line 86: I pulled the meta logic out into this final step instead of having it sprinkled through every branch of the if/else tree

PR #306, opened 2020-07-14
Fix ghost message bug from pass-by-reference

christianbundy:
Problem: There's an issue where sometimes streams don't properly unbox
messages if you close the server and restart it, which is very
confusing. I've walked through the entire stack until I ended up here.

Solution: Add a test so that we can start putting together solutions to
this problem.

---

Problem: There's an issue where sometimes streams don't properly unbox
messages if you close the server and restart it, which is very
confusing. I've walked through the entire stack until I ended up here.

Solution: Add a test so that we can start putting together solutions to
this problem.

### DIFF
--- a/autobox.js
+++ b/autobox.js
@@ -1,6 +1,6 @@
 const HLRU = require('hashlru')
+const cloneDeep = require('lodash.clonedeep')
 const { metaBackup } = require('./util')
-const { cloneDeep } = require('lodash')
 
 function isFunction (f) { return typeof f === 'function' }
 function isString (s) { return typeof s === 'string' }
@@ -94,24 +94,11 @@ const CACHE_SIZE = 512
 
 function unboxWithCache (id) {
   var cache = HLRU(CACHE_SIZE)
-  // console.log('instantiating unbox.withCache', id)
 
-  // var state = {
-  //   total: 0,
-  //   cachedUnbox: 0,
-  //   unbox: 0
-  // }
   function cachedUnbox (msg, readKey, unboxers) {
-    // state.total++
-    // if (state.total % 1000 === 0) {
-    //   const percent = Math.floor(state.cachedUnbox / (state.cachedUnbox + state.unbox) * 1e3) / 1e2
-    //   console.log(`cache saves: ${percent}% (${state.cachedUnbox} / ${state.cachedUnbox + state.unbox})`)
-    // }
     if (!msg || !isString(msg.value.content)) return msg
 
     const cached = cache.get(msg.key)
-    // if (cached !== undefined) state.cachedUnbox++
-    // else state.unbox++
 
     if (cached === false && !readKey) return msg
     else if (cached) return cached
--- a/autobox.js
+++ b/autobox.js
@@ -1,5 +1,6 @@
 const HLRU = require('hashlru')
 const { metaBackup } = require('./util')
+const { cloneDeep } = require('lodash')
 
 function isFunction (f) { return typeof f === 'function' }
 function isString (s) { return typeof s === 'string' }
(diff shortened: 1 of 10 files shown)


autobox.js
```diff
@@ -1,5 +1,6 @@
 const HLRU = require('hashlru')
 const { metaBackup } = require('./util')
+const { cloneDeep } = require('lodash')
```
mixmix, line 3: this is heavy (pulls whole of lodash in).

instead do this:

```js
const cloneDeep = require('lodash.clonedeep')
```

test/ghost.js
```diff
@@ -0,0 +1,81 @@
+const tape = require('tape')
+const createSsb = require('./util/create-ssb')
+const ssbTribes = require('ssb-tribes')
+const ssbBacklinks = require('ssb-backlinks')
+
+const plugins = [ssbBacklinks, ssbTribes]
```
mixmix: this file + tests needs more descriptive names

autobox.js
mixmix: DONE

test/ghost.js
mixmix: DONE

test/unbox-with-cache.js
```diff
@@ -0,0 +1,88 @@
+/*
+  This file exists because we were seeing strange behaviour around indexing
+  of encrypted messages.
+  The bug was tracked down to the unbox cache in autobox.js, which was storing
+  a reference to a state which was being mutated elsewhere.
+  These tests are left to guard against this bug recurring
+*/
```
mixmix, line 7: renamed this file, and left this bit of history here

test/unbox-with-cache.js
```diff
@@ -0,0 +1,88 @@
...
+]
+const createSsb = require('./util/create-ssb')
+
+// This test uses get() from SSB-DB to unbox the published message.
+// It seems to work fine, and `msg.value.content` is always an object.
+tape('unbox.withCache - async', (t) => {
+  const name = `ghost-get-${Date.now()}`
+  const a = createSsb(name, { temp: false }, plugins)
+
+  const asyncGet = (node, key, cb) => node.get({ id: key, private: true, meta: true }, cb)
```
mixmix, line 23: I love this pattern. God I hate writing all that junk in get calls

mixmix:
ok windows... (travis)

autobox.js
```diff
@@ -93,31 +94,18 @@ const CACHE_SIZE = 512
...
 
     if (cached === false && !readKey) return msg
     else if (cached) return cached
 
     const result = unbox(msg, readKey, unboxers)
     if (isString(result.value.content)) cache.set(msg.key, false)
-    else cache.set(msg.key, result)
+    else cache.set(msg.key, cloneDeep(result))
 
     return result
```
mixmix, line 110: So I'm still trying to understand what was going wrong with the streams.
Was it that later something is using the result and mutating the result (e.g. by running some `originalValue` or similar) which mutates the cached value?

That I can understand. What I can't understand is why restarting the server is what generates that state when you run a query which was previously running fine ... doesn't a server restart clear this LRU cache?
I see it passes the `ssb-tribes` test though ... so that's good.

christianbundy:
> That I can understand. What I can't understand is why restarting the server is what generates that state when you run a query which was previously running fine ... doesn't a server restart clear this LRU cache?

The cache is being cleared, but **completely unrelated queries** apparently mutate the data. In this case, the [`createUserStream()`](https://github.com/ssbc/ssb-tribes/blob/cccd2b9c804d3a50f5f92c2aed845ee8940db9c0/listen.js#L28) doesn't have `{ private: true }`, which means the message is passed to `originalValue()`, and apparently **this mutates others reference to the message**. Here's a horrifying test that proves it.

Layers of race conditions + layers of shared mutable state = :scream: 

I've written a patch (bfd4cea31a6c90b2262b6dcaa5beb2e79ea0276b) to ensure that `originalValue()` does a copy-by-value to avoid this problem throughout the stack, but I'm pretty mortified that it was happening in the first place. The silver lining is that our mutation was applying the boxer, not unboxing, so this bug is/was confusing and time-consuming rather than outright dangerous.

PR #307, opened 2020-07-15
Fix pass-by-reference from cache

christianbundy:
Problem: HashLRU passes references, not values, so when we use the cache
we're accidentally creating a scenario where one variable being mutated
can change the contents of the results of another query.

Solution: Use `lodash.cloneDeep()` on the results from the HashLRU cache
to ensure that we're always passing by value rather than by reference.

---

I was going to make an issue for this, but I figured I should just fix it instead. I think this is better than `Object.freeze()` because that makes you *think* that you can change an object (no errors) but quietly drops any changes that you try to make.

### DIFF
--- a/autobox.js
+++ b/autobox.js
@@ -1,5 +1,6 @@
 const HLRU = require('hashlru')
 const { metaBackup } = require('./util')
+const cloneDeep = require('lodash.clonedeep')
 
 function isFunction (f) { return typeof f === 'function' }
 function isString (s) { return typeof s === 'string' }
@@ -100,7 +101,7 @@ function unboxWithCache (id) {
     const cached = cache.get(msg.key)
 
     if (cached === false && !readKey) return msg
-    else if (cached) return cached
+    else if (cached) return cloneDeep(cached)
 
     const result = unbox(msg, readKey, unboxers)
     if (isString(result.value.content)) cache.set(msg.key, false)
--- a/test/unbox-with-cache.js
+++ b/test/unbox-with-cache.js
@@ -88,7 +88,7 @@ tape('unbox.withCache - source', (t) => {
 // query. This was written to illustrate a problem where `unboxValue()` would
 // **mutate the results of other queries** and re-box messages that were meant
 // to be private.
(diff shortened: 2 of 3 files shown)


test/unbox-with-cache.js
```diff
@@ -111,3 +111,27 @@ tape('shared mutable state (source)', (t) => {
...
+    pull(
+      ssb.createUserStream({ id: ssb.id, reverse: true, limit: 1, private: true }),
+      pull.collect((err, [privateMessageA]) => {
+        t.error(err)
+        const copy = cloneDeep(privateMessageA)
+        pull(
+          ssb.createUserStream({ id: ssb.id, reverse: true, limit: 1, private: true }),
+          pull.collect((err, [privateMessageB]) => {
+            privateMessageA.value = 'this should not effect privateMessageB'
+            t.deepEqual(copy, privateMessageB, 'active tampering should not mutate variables')
```
mixmix, line 130: further we can check they don't have the same reference I think

mixmix: this test is a little unclear to me. I don't know what passive / active means.

Also
Think you could check `copy` against `privateMessageA` later, should be the same (deepEqual).
I think you should check `privateMessageA` against `privateMessageB` - we expect deepEqual ... what do we expect about equal ..should the references be the same?

autobox.js
```diff
@@ -100,7 +101,7 @@ function unboxWithCache (id) {
     const cached = cache.get(msg.key)
 
     if (cached === false && !readKey) return msg
-    else if (cached) return cached
+    else if (cached) return cloneDeep(cached)
```
mixmix, line 104: ah yes I see, I think this is quite the right place to put this... on the exit ensure _no-one_ can mutate the cache

mixmix: :heart:

test/unbox-with-cache.js
mixmix: Oh ignore me, reading tests in editor they make complete sense.
I just modified the test titles to be clearer (I think)

PR #308, opened 2020-07-22
Fix database bug when bulk appending messages

christianbundy:
Problem: If you publish more than 1000 messages at once, at least one
message calls back without a message key.

Solution: Add a failing test to make the problem easier to reproduce.

cc: @mixmix @arj03 @staltz


christianbundy:
Hmm:

https://github.com/dominictarr/async-write/blob/db5574a35003f690dad73ff28d54b8692afbc14a/index.js#L19-L21

mixmix:
Some context (in case missing) we're wanting to import a csv of family
records and want to create a lot of profiles at once for family trees.

staltz:
Sounds like a good discovery, but would we merge a failing test without merging its solution?

And to address the use case: instead of writing many synchronously, couldn't you wait for the response for writing msg N before requesting the write for msg N+1? In pure callback JS this is annoying to write, but with some async helper functions (maybe even pull-stream) this could be written elegantly.

test/flush-bug.js
```diff
@@ -0,0 +1,19 @@
+const tape = require('tape')
+const createSsb = require('./util/create-ssb')
+
+// If you write faster than the database can save messages, it adds them to
+// some kind of queue, which eventually gets 'flushed'. Unfortunately it looks
+// like when you publish more than 1000 messages very quickly, the 1000th
+// message comes back as `undefined` rather than `{ key, value }` like it
+// should.
+tape('test ', (t) => {
```
staltz, line 9: There could be a better test title here.

christianbundy:
> Sounds like a good discovery, but would we merge a failing test without merging its solution?

Oops, no -- this was meant to be a draft.

> And to address the use case: instead of writing many synchronously, couldn't you wait for the response for writing msg N before requesting the write for msg N+1?

Yep, that would be a fine workaround. I'm able to work around the bug, but I wanted to file an issue to illustrate the problem so we can eventually fix it.

minimal.js
```diff
@@ -83,36 +82,50 @@ module.exports = function (dirname, keys, opts) {
...
+
+  const queue = (message, cb) => {
+    try {
+      V.append(state, hmacKey, message)
+      cb(null, state.queue[state.queue.length - 1])
+      if (writing === false) {
+        write()
+      }
+    } catch (e) {
+      cb(e)
```
christianbundy, line 124: Previously this `try...catch` was happening inside of Async-Write, I promise I didn't add it for fun.

minimal.js
```diff
@@ -128,20 +141,14 @@ module.exports = function (dirname, keys, opts) {
...
-    queue(msg, function (err) {
+    queue(msg, function (err, message) {
       if (err) return cb(err)
-      var data = state.queue[state.queue.length - 1]
-      flush.add(() => cb(null, data))
+      flush.add(() => cb(null, message))
     })
   }))
 
-  db.buffer = function buffer () {
```
christianbundy, line 138: Removed because it was a shortcut method to introspect Async-Write and wasn't exported or used anywhere.

christianbundy:
Interesting, tests are passing locally but Travis is seeing a regression in `test/del.js`. I've been wondering if the 100 ms timer from the old code was masking a race condition, maybe that's what's happening.

staltz:
Took a big-picture look at each commit. I still need to look carefully into the heart of the main solution here, but so far I could say that you could try to squash some commits together, e.g. the commit that removes `throw new Error('wtf')` could be blended together with the commit that introduced it in the first place. It makes it easier to debug (`git blame` or file history) later on when people are going back in time, navigating the commits.

christianbundy:
> I still need to look carefully into the heart of the main solution here

Thanks, I'll wait for that.

> you could try to squash some commits together

I think we should probably outline some sort of SSBC policy around squashing. I'd prefer to have the real development history instead of a fictional one, but I'm fine just squashing [after this branch has been reviewed] and discussing a policy later. I'd need to squash #310 and force-push there too, since it depends on this branch.

minimal.js
```diff
@@ -128,20 +141,14 @@ module.exports = function (dirname, keys, opts) {
       return cb(err)
     }
 
-    queue(msg, function (err) {
+    queue(msg, function (err, message) {
```
mixmix, line 144: I think there's a behaviour change here. Your `queue` calls back with _every_ callback passed to it.
AsyncWrite runs the callback when:
1. (buffer is full)
   - a), and (not already writing)
   - b)  and (already writing) AND **(is was the latest callback added)**
2. (buffer is not full)

Path 1 (b) looks like it basically swallows all but the last callback from running  once the buffer was already full

mixmix: https://github.com/dominictarr/async-write/blob/master/index.js#L36

mixmix: I _think_ that could be the bug which was returning undefined after you top 1000 messages

minimal.js
```diff
@@ -83,36 +82,50 @@ module.exports = function (dirname, keys, opts) {
...
-  }, 100)
-
-  queue.onDrain = function onDrain () {
-    if (state.queue.length === 0) {
-      flush.runAll()
+  }
+
+  const queue = (message, cb) => {
+    try {
+      V.append(state, hmacKey, message)
```
mixmix, line 118: I think I'd like a note which tells readers that `V.append` mutates `state.queue`. I know it's not your code, but it would make maintenance + readability easier

minimal.js
```diff
@@ -83,36 +82,50 @@ module.exports = function (dirname, keys, opts) {
...
   var append = db.rawAppend = db.append
   db.post = Obv()
-  var queue = AsyncWrite(function (_, cb) {
-    var batch = state.queue
-    state.queue = []
-    append(batch, function (err, v) {
+
+  let writing = false
+
+  const write = () => {
```
mixmix, line 88: :fire: 

I would rather have `if (writing) return` at the top of this to defend against accidental multiple calls. 

You can then replace `if (writing === false) write()` with just `write()` below

mixmix: `<style>` I think I'd also be tempted to staple the isWriting variable to this function like `write.isRunning`

minimal.js
```diff
@@ -83,36 +82,50 @@ module.exports = function (dirname, keys, opts) {
...
-  }, function (_state) {
-    return state.queue.length > 1000
-  }, function isEmpty (_state) {
-    return !state.queue.length
-  }, 100)
-
-  queue.onDrain = function onDrain () {
-    if (state.queue.length === 0) {
-      flush.runAll()
+  }
```
mixmix, line 114: :question: behaviour change - your queue no longer has a timer, as in it doesn't wait a minimum time between each write call. I don't know if that was needed / if there are any implications of removing.
I don't think it's dangerous though

test/del.js
```diff
@@ -19,23 +20,25 @@ function run (opts = {}) {
...
+              pull.drain(() => {
+                t.fail('no messages should be available')
+              }, () => {
+                ssb.get(msg.key, (err) => {
+                  t.ok(err)
+                  t.equal(err.code, 'flumelog:deleted')
+                  ssb.close(err => {
+                    t.error(err, 'ssb.close - del (delete message)')
+                    t.end()
+                  })
```
mixmix, line 38: Christian said : 

> the change in code exposed a race condition in this test, this just fixes that

minimal.js
christianbundy: Yeah, it looks like while Async-Write is writing it drops all callbacks except the most recent one added. So if your buffer is full after 1000 messages and you supply 2000, it'll drop callbacks 1001-1999.

The weird thing is that the bug wasn't that the callbacks weren't running, it's that they *were* running but `msg === undefined`. I don't really understand why.

minimal.js
christianbundy: > I would rather have `if (writing) return` at the top of this to defend against accidental multiple calls.

I thought so too, but then `write()` can't call itself without setting `writing = false`. If you do that then both `queue()` and `write()` both call `write()`, and although only one of them succeeds you still get duplicate function calls. Maybe you could propose a patch that you think would do this better?

> I think I'd also be tempted to staple the isWriting variable to this function like `write.isRunning`

I'd be comfortable seeing this implemented as a 'writer' class that holds the SSB-Validate state and writing state internally, but I have a much harder time understanding code that staples properties onto random variables that aren't objects.

minimal.js
christianbundy: I think that came from Async-Write's goal to "keep a buffer ready to write at all times", whereas this code has the goal to **immediately write all queued messages to the database**. I've placed the queue limit and timers off to the side in my YAGNI bucket until we can write some tests or benchmarks that show why they're useful.

minimal.js
mixmix: this makes sense, happy to leave as is.
Would be nice to collect state together as much as possible, perhaps that's a different patch though

minimal.js
mixmix: Shall we leave an issue on the repo with notes about what was changed as an FYI for future spelunkers?

minimal.js
mixmix: <3

minimal.js
mixmix: :+1:

mixmix:
Merged. Will publish shortly along with other changes and leave notes on the Release tag

mixmix:
Great work @christianbundy , thank you <3

Issue #300, opened 2020-06-18
Intermittent test failures on Windows (Travis CI)


christianbundy:
Most of the time Travis builds work great, but every few builds we have a failure on Windows that requires restarting the test. It's not a show-stopper, but it's likely indicative of a deeper problem.

christianbundy:
Example:

```
not ok 241 Error [ERR_SERVER_NOT_RUNNING]: Server is not running.

  ---

    operator: error

    at: <anonymous> (C:\Users\travis\build\ssbc\ssb-db\node_modules\secret-stack\core.js:258:21)

    stack: |-

      Error [ERR_SERVER_NOT_RUNNING]: Server is not running.

          at Server.close (net.js:1602:12)

          at Object.onceWrapper (events.js:421:28)

          at Server.emit (events.js:315:20)

          at emitCloseNT (net.js:1655:8)

          at processTicksAndRejections (internal/process/task_queues.js:83:21)

  ...

not ok 242 plan != count

  ---

    operator: fail

    expected: 5

    actual:   6

    stack: |-

      Error: plan != count

          at Test.assert [as _assert] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:260:54)

          at Test.bound [as _assert] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:84:32)

          at Test.fail (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:354:10)

          at Test.bound [as fail] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:84:32)

          at Test.assert [as _assert] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:346:14)

          at Test.bound [as _assert] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:84:32)

          at Test.error (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:407:10)

          at Test.bound [as ifError] (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:84:32)

          at Test.end (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:173:14)

          at bound (C:\Users\travis\build\ssbc\ssb-db\node_modules\tape\lib\test.js:84:32)

  ...

[...]

PR #310, opened 2020-07-24
Fix intermittent Windows test failures

christianbundy:
Problem: For some reason the Travis CI Windows builds fail
intermittently with a specific error that doesn't make any sense to me.
We've been just restarting the build, since the erorr (can't close the
server because the server is already closed) doesn't matter, but it's
still very obnoxious.

Solution: Ignore the error entirely and don't worry about it. I don't
think that we care if the server is already closed, the important thing
is that it's closed. I'm okay with `close()` being idempotent.

Resolves: #300

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,6 +83,10 @@ module.exports = {
           _close((err) => {
             setImmediate(() => {
               if (typeof cb === 'function') {
+                if (typeof err === 'object' && err !== null && err.code === 'ERR_SERVER_NOT_RUNNING') {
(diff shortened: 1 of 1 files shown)


christianbundy:
:(

christianbundy:
FIXED!!!

christianbundy:
(This pulls in changes from #308 because that branch fixed a `Date.now` being used as a filename when it should've been `Date.now()`)

mixmix:
This is a great find, will make maintenance a lot more joyful <3

PR #312, opened 2020-07-31
Pass feed state to boxer

christianbundy:
Alternative to #311 which automatically passes feed state to the boxer, exposing a method instead of an observable for further use.

---

Problem: Some boxers, like SSB-Tribe, depend on the feed state. We've
tried using hacks to query the database before publishing a message, but
we really just need to be passed the feed state (according to the
validator) each time the boxer is asked to box a message.

Solution: Pass the relevant feed state from the validator to the boxer.
There are also some circumstances where we want some specific feed
state, like when boxing a message to pass to `add()` rather than
`append()`, and so I've added a `getFeedState()` method that returns
those results.

### DIFF
--- a/autobox.js
+++ b/autobox.js
@@ -5,7 +5,7 @@ const cloneDeep = require('lodash.clonedeep')
 function isFunction (f) { return typeof f === 'function' }
 function isString (s) { return typeof s === 'string' }
 
-function box (content, boxers) {
+function box (content, boxers, feedState) {
   if (!content.recps) return content
 
   if (typeof content.recps === 'string') content.recps = [content.recps]
@@ -15,7 +15,7 @@ function box (content, boxers) {
   var ciphertext
   for (var i = 0; i < boxers.length; i++) {
     const boxer = boxers[i]
-    ciphertext = boxer(content)
+    ciphertext = boxer(content, feedState)
 
     if (ciphertext) break
   }
--- a/index.js
+++ b/index.js
@@ -31,6 +31,7 @@ var manifest = {
   getVectorClock: 'async',
   help: 'sync',
   latest: 'source',
+  getFeedState: 'async',
   latestSequence: 'async',
   links: 'source',
   messagesByType: 'source',
@@ -154,6 +155,7 @@ module.exports = {
       since                    : since,
 
       latest                   : ssb.latest,
+      getFeedState             : valid.async(ssb.getFeedState, 'feedId'),
       getLatest                : valid.async(ssb.getLatest, 'feedId'),
       latestSequence           : valid.async(ssb.latestSequence, 'feedId'),
       createFeed               : ssb.createFeed,
--- a/minimal.js
+++ b/minimal.js
@@ -128,9 +128,25 @@ module.exports = function (dirname, keys, opts) {
 
   db.queue = waitForValidators(queue)
 
+  const getFeedState = (feedId) => {
+    const feedState = state.feeds[feedId]
+    if (!feedState) return { id: null, sequence: 0 }
+    // NOTE this covers the case where you have a brand new feed (or new createFeed)
+
+    // Remove vestigial properties like 'timestamp'
+    return {
+      id: feedState.id,
+      sequence: feedState.sequence
+    }
+  }
+  db.getFeedState = waitForValidators((feedId, cb) => {
+    cb(null, getFeedState(feedId))
+  })
+
   db.append = waitForBoxers(waitForValidators(function dbAppend (opts, cb) {
     try {
-      const content = box(opts.content, boxers)
+      const feedState = getFeedState(opts.keys.id)
+      const content = box(opts.content, boxers, feedState)
       var msg = V.create(
         state.feeds[opts.keys.id],
         opts.keys,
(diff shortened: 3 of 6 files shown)


autobox.js
```diff
@@ -15,7 +15,7 @@ function box (content, boxers) {
   var ciphertext
   for (var i = 0; i < boxers.length; i++) {
     const boxer = boxers[i]
-    ciphertext = boxer(content)
+    ciphertext = boxer(content, feedState)
```
mixmix, line 18: we need to document this change in `addBoxer` API

index.js
```diff
@@ -27,6 +27,7 @@ var manifest = {
   createWriteStream: 'sink',
   del: 'async',
   get: 'async',
+  getFeedState: 'async',
```
mixmix, line 30: If I'm correct this exposes this over muxrpc. Do we want to or need to do this?
If we don't absolutely need to I would rather not, because it's harder to take back than to add later

mixmix:
YES, I'm massively a fan of this over #311 

Closing that. There's a couple of pending questions, will merge this once resolved

minimal.js
```diff
@@ -114,9 +114,26 @@ module.exports = function (dirname, keys, opts) {
     })
   })
 
+  const formatFeedState = (feedState) => {
+    if (feedState) {
```
mixmix, line 118: :fire: - I don't know how the `feedState` is maintained ... but I wonder if there's a chance where it could be loading and this comes through as `undefined` ... I think I want to check some of that before merging this bit

autobox.js
mixmix: My bad, you already did!

index.js
mixmix: Have disabled

test/get-feed-state.js
```diff
@@ -0,0 +1,33 @@
...
+  t.deepEqual(initialState, { id: null, sequence: 0 }, 'empty feed')
+
+  await publish({ type: 'boop' })
+
+  var laterState = await getFeedState(ssb.keys.id)
+  t.equal(laterState.sequence, 1, 'sequence increases after publish')
+  t.true(isMsg(laterState.id), 'msgId present after publish')
+
+  var otherState = await getFeedState(otherKey)
+  t.deepEqual(otherState, null, 'returns null for remote feeds it does not know about')
```
mixmix, line 23: @christianbundy I changed this behaviour. I think it was dangerous to return a potentially erroneous state when one is not actually known

minimal.js
```diff
@@ -212,6 +228,14 @@ module.exports = function (dirname, keys, opts) {
...
         }
       }
+      if (!state.feeds[keys.id]) {
+        state.feeds[keys.id] = {
+          id: null,
+          timestamp: null,
+          sequence: 0,
+          queue: []
+        }
+      }
```
mixmix, line 238: @christianbundy I added this to the initialisation of validation state, it is the replacement of your case where you're looking for the state of your empty feed ... this is the one case where we know for sure that a feed is empty

minimal.js
christianbundy: All of the functions that call `getFeedState()` are wrapped with `waitForValidators()`, so SSB-Validate must be ready.

minimal.js
christianbundy: I think this makes the behavior inconsistent, since `getFeedState()` is meant to return the state of any feed passed. After this change it returns different results for an empty feed depending on whether it's your feed ID.

index.js
christianbundy: My policy is to expose everything over MuxRPC that's exported by the JS module unless it doesn't work over the network (e.g. `_flumeUse()`). What's the benefit to maintaining two similar-but-different APIs here?

minimal.js
mixmix: yes but your feed is the only one you can authoritatively know is empty.
If you ask for the state of another feed and it's not in the `db.last` index all you can say is "I don't know what the state is". That could mean "it's empty" or "I haven't got it yet". Returning a result which suggests you know it's empty is dangerous IMO

mixmix: discussed with Christian on phone. 
we decided to remove the feedId arg from the method, and to add this method back to the manifest

mixmix: i.e. it will only get your current feed state

mixmix:
ooops, my rebase squashed this kinda a little more aggressively than intended ... nevertheless, looks fine

test/append.js
```diff
@@ -121,23 +121,36 @@ tape('append (read back)', function (t) {
...
-  const assertsPerMessage = 4;
-  t.plan(messages * assertsPerMessage);
+  const assertsPerMessage = 4
+  const plan = messages * assertsPerMessage
+  var pass = 0
+
+  function testEqual (a, b) {
+    if (a !== b) {
+      process.stdout.write('\n')
+      t.equal(a, b)
```
mixmix, line 137: this just quietens down this particular test so that if there's an error you see it, but otherwise it doesn't log heaps so you can feasibly scroll past this test in the termina

christianbundy: Alternatively:

```suggestion
    if (a !== b) {
      t.fail('messages must be equal')
```

PR #319, opened 2021-01-03
db.get to return flumedb offset in the 3rd arg

staltz:
Context: https://github.com/ssb-ngi-pointer/ssb-db2/pull/97

Problem: in ssb-db2 we need to efficiently perform migration from log.offset to log.bipf, and for that we need to retrieve the *offset on log.offset of the latest migrated record*, because our process works like this: (1) pick the last record in log.bipf, (2) pick its `msg.key`, (3) retrieve the record on log.offset that matches `msg.key`, (4) retrieve the offset of that record on log.offset, (5) continue scanning log.offset from that specific offset.

Solution: to perform steps (4) and (5), we need `get(msgkey, cb)` to give us also the offset. This PR returns the offset as the 3rd CB argument. Why the 3rd arg? Because it's backwards compatible with whatever ssb-db was doing and providing to other parts of the stack (i.e. we don't need to change the second arg, which would be a breaking change), **and** because flumedb already returns the offset as the 3rd arg, it would be natural to propagate that.

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -2,6 +2,7 @@
 var tape = require('tape')
 var pull = require('pull-stream')
 var crypto = require('crypto')
+var util = require('util')
 
 var createSSB = require('./util/create-ssb')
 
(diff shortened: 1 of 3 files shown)


arj03:
Looks good to me, it needs an update to README though. As this is backwards compatible I don't see a big problem risk.

staltz:
Added a commit to update README too.

mixmix (commented):
Sleepy 4am review :
Does this in any way mess up `const msg = await promisify(ssb.get)(key)`

There might be code written like this

staltz:
@mixmix good question. This change would have no breaking change if you're using pify or promisify-tuple or promisify-4loc, the resolved promise will have the same value as before, i.e. `msg`, and the `offset` would not surface (same behavior as before).

mixmix:
Write a test I reckon

staltz:
Added tests

test/add.js
```diff
@@ -45,10 +46,41 @@ function run (opts) {
...
+      })
+    })
+  })
+
+  tape('get works with promisify', function (t) {
+    ssb.publish({ type: 'okay' }, function (err, msg) {
+      t.error(err)
+      t.equal(msg.value.content.type, 'okay')
+
+      util.promisify(ssb.get)(msg.key).then(msgVal => {
```
mixmix, line 73: Cool! I'm happy for this to be merged then

mixmix (approved):

staltz:
Thanks @mixmix and @arj03 for reviewing!

PR #321, opened 2021-09-14
wait till rebuild is done before closing db

mixmix:
**Context**
- there's a really nasty error you can get where if you call `ssb.close` while a db rebuild is underway you get a scary "segmentation fault" error sometimes. It isn't common, but it often breaks tests in an intermittent way

**Solution**
- I've made the `close` method wait till `rebuild` is done before proceeding

I currently have a hack on `scuttle-testbot` which basically hooks this same thing into place but I was scared that we might hit it in production and bad things would happen.

A nicer solution *could* be digging deeper into the stack, but I don't have the time to do that.
An alternative solution would be to log out a helpful error and then let the segmentation fault happen


index.js
```diff
@@ -72,16 +72,22 @@ module.exports = {
     var ssb = create(opts.path, opts, opts.keys)
     //treat the main feed as remote, because it's likely handled like that by others.
     var feed = ssb.createFeed(opts.keys, {remote: true})
-    var _close = api.close
```
staltz, line 75: +1 to getting rid of underscore names

staltz (approved):

staltz:
> "segmentation fault"

Oooh, I had this once and thought "what on Earth, in Node.js??" and it must have been this issue.

PR #327, opened 2024-07-13
fix: avoid overflow error when computing overall progress

black-puppydog:
Hello. In good old ssb fashion I'm necro-PR'ing this. :)

This is a fix for a lock-up for the user if the offset log goes beyond 2^31 bytes in length. Since the `~~` uses a (signed) int32 representation, it produces a negative number in these cases. Result: the `progress.indexes.current` value is negative, thus smaller than the `progress.indexes.target` value, and Patchwork refuses to produce any further messages.


soapdog:
I don't have access to ssbc org anymore. Can someone merge this please?

